PR #4, opened 2020-02-26
Shared modules

ojemuyiwa:
Shared library implemented
Analytics module configured
Environments Interface set-up

#TISNEW-3932

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,12 +3,13 @@
   "version": "0.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "ng serve",
+    "start": "npm run build:shared && ng serve",
     "build": "ng build",
     "test": "ng test",
     "lint": "ng lint && npm run lint:styles",
     "lint:styles": "stylelint \"src/**/*.scss\"",
-    "e2e": "ng e2e"
+    "e2e": "ng e2e",
+    "build:shared": "ng build --project=hee-shared"
   },
   "private": true,
   "dependencies": {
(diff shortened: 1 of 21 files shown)


tsconfig.json
```diff
@@ -17,10 +17,19 @@
     "lib": [
       "es2018",
       "dom"
-    ]
+    ],
+    "paths": {
+      "hee-shared": [
+        "dist/hee-shared/hee-shared",
```
omarCreativeDev: is this path correct?

package.json
```diff
@@ -21,13 +21,15 @@
...
     "@angular/service-worker": "~9.0.2",
+    "angulartics2": "^8.3.0",
     "nhsuk-frontend": "^3.0.3",
     "rxjs": "~6.5.4",
     "tslib": "^1.10.0",
     "zone.js": "~0.10.2"
   },
   "devDependencies": {
     "@angular-devkit/build-angular": "~0.900.3",
+    "@angular-devkit/build-ng-packagr": "~0.900.3",
```
omarCreativeDev: as discussed it will be good if you can update the ng start and build scripts to reflect the shared modules consumption

tsconfig.json
ojemuyiwa: yes its the build path for compilation/tree shaking

package.json
ojemuyiwa: excellent. will do

tsconfig.json
omarCreativeDev: cool

PR #5, opened 2020-03-02
3rd party tools

ojemuyiwa:
Configured hotjar for production and sonarcloud

#TISNEW-3933

### DIFF
--- a/projects/hee-shared/src/lib/hotjar/hotjar.module.ts
+++ b/projects/hee-shared/src/lib/hotjar/hotjar.module.ts
@@ -1,0 +1,52 @@
+import { NgModule, InjectionToken, Inject, Optional, SkipSelf, ModuleWithProviders } from '@angular/core';
+import { InjectScript } from '../utilities.functions';
+import { DOCUMENT } from '@angular/common';
+
+export interface HotjarConfig {
+  hotjarid: number;
+  hotjarsv: number;
+  enabled: boolean;
+}
+
+const HotjarConfigValue = new InjectionToken<HotjarConfig>('HotjarConfig');
+
+@NgModule({
+  declarations: []
+})
+export class HotjarModule {
+  private hotjarscript: HTMLScriptElement;
+
+  constructor(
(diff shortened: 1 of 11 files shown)


projects/hee-shared/src/lib/hotjar/hotjar.module.ts
```diff
@@ -0,0 +1,52 @@
+import { NgModule, InjectionToken, Inject, Optional, SkipSelf, ModuleWithProviders } from '@angular/core';
+import { InjectScript } from '../utilities.functions';
+import { DOCUMENT } from '@angular/common';
+
+export interface HotjarConfig {
+  hotjarid: number;
```
omarCreativeDev: is it possible to camelcase this interface?

projects/hee-shared/src/lib/hotjar/hotjar.module.ts
```diff
@@ -0,0 +1,52 @@
...
+  enabled: boolean;
+}
+
+const HotjarConfigValue = new InjectionToken<HotjarConfig>('HotjarConfig');
+
+@NgModule({
+  declarations: []
+})
+export class HotjarModule {
+  private hotjarscript: HTMLScriptElement;
```
omarCreativeDev: is it possible to camelcase this property?

projects/hee-shared/src/lib/hotjar/hotjar.module.ts
ojemuyiwa: Revised all and camel cased across using angular style guide: https://angular.io/guide/styleguide

omarCreativeDev:
pr looks good but seems like prettifier hasn't run on your commits. are you using source tree to commit and push? if so not sure how husky would come in to play to auto prettify modified files and also run js and css linting. maybe there is a way to configure source tree to use husky or auto prettify and lint files?

or alternatively if you commit via the command line everything will be automagically done for you as husky pre-commit hook will prettify and run linting before allowing to commit

PR #6, opened 2020-03-02
fix unit tests

omarCreativeDev:
- fix broken unit tests
- remove empty scss file
- converted non arrow function in `hee-shared/src/lib/analytics/analytics.module.ts:80:23` as wasn't able to commit

### DIFF
--- a/projects/hee-shared/src/lib/analytics/analytics.module.ts
+++ b/projects/hee-shared/src/lib/analytics/analytics.module.ts
@@ -92,9 +92,9 @@ export class AnalyticsModule {
     if (!this.gascript && typeof this.win.gtag === "undefined") {
       this.gascript = true;
       this.win.dataLayer = this.win.dataLayer || [];
-      this.win.gtag = () => {
-        (window as any).dataLayer.push(arguments);
-      }; // The 'arguments' object cannot be referenced in an arrow function in ES3 and ES5. Consider using a standard function expression.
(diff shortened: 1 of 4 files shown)


projects/hee-shared/src/lib/analytics/analytics.module.ts
```diff
@@ -74,11 +89,13 @@ export class AnalyticsModule {
...
   private initializeGtag(): void {
-    if (!this.gascript && typeof this.win.gtag === 'undefined') {
+    if (!this.gascript && typeof this.win.gtag === "undefined") {
       this.gascript = true;
       this.win.dataLayer = this.win.dataLayer || [];
-      this.win.gtag = function() { (window as any).dataLayer.push(arguments); }; // The 'arguments' object cannot be referenced in an arrow function in ES3 and ES5. Consider using a standard function expression.
-      this.win.gtag('js', new Date());
+      this.win.gtag = () => {
+        (window as any).dataLayer.push(arguments);
+      }; // The 'arguments' object cannot be referenced in an arrow function in ES3 and ES5. Consider using a standard function expression.
```
ojemuyiwa: comment on this line was because I didn't wanna break the google snippet. Not sure if this works / compiles

omarCreativeDev: ah yes you are correct. just pushed a fix

PR #17, opened 2020-03-09
Tisnew 4044 page not found route

omarCreativeDev:
Tisnew 4044 page not found route


omarCreativeDev:
> may revise to errormodule with different routes server errors etc including page not found. depending on routing strategy

error module sounds like a good idea! something to pick up when we tackle server errors. good shout

PR #29, opened 2020-03-27
refactor: fe updates due to api changes

omarCreativeDev:
TISNEW-4088

- refactor fe code in order to align with be endpoint schema
- added new local environment file alongside new `npm start:local` script
- updated proxy config
- update unit and e2e tests
- upgrade prettier & cypress
- cleanup npm scripts


angular.json
```diff
@@ -103,6 +101,20 @@
               ],
               "serviceWorker": true,
               "ngswConfigPath": "ngsw-config.json"
+            },
+            "local": {
```
ojemuyiwa: dev is environment.ts, staging environment.stage.ts, and prod. which environment is local

omarCreativeDev: yea thats correct. local is for when running the be locally on your own machine

ojemuyiwa: yeah thats dev

ojemuyiwa: with debug zones enabled

omarCreativeDev: dev `environment.ts` is for consuming FE controlled mock api's with `mocky.io` during local development.

I added a new local environment file `environment.local.ts` so provide the ability to consume locally running docker based BE during local development

- so as before to spin up app for development with mocky `npm run start`
- and to spin up app for development with local BE `npm run start:local`

PR #30, opened 2020-03-30
Tisnew 4109

ojemuyiwa:
PR for history view

### DIFF
--- a/src/app/dashboard/dashboard.module.ts
+++ b/src/app/dashboard/dashboard.module.ts
@@ -6,9 +6,20 @@ import { DashboardRoutingModule } from "./dashboard-routing.module";
 import { UnderNoticeState } from "./under-notice/state/under-notice.state";
 import { UnderNoticeComponent } from "./under-notice/under-notice.component";
 import { MaterialModule } from "../shared/material/material.module";
+import { TraineeSummaryComponent } from "./trainee-summary/trainee-summary.component";
+import { RevalidationHistoryComponent } from "./revalidation-history/revalidation-history.component";
+import { RevalidationNotesComponent } from "./revalidation-notes/revalidation-notes.component";
+import { RevalidationFormComponent } from "./revalidation-form/revalidation-form.component";
 
 @NgModule({
-  declarations: [DashboardComponent, UnderNoticeComponent],
+  declarations: [
+    DashboardComponent,
+    UnderNoticeComponent,
+    TraineeSummaryComponent,
+    RevalidationHistoryComponent,
+    RevalidationNotesComponent,
+    RevalidationFormComponent
+  ],
   imports: [
     MaterialModule,
     SharedModule,
--- a/src/app/dashboard/revalidation-form/revalidation-form.component.scss
+++ b/src/app/dashboard/revalidation-form/revalidation-form.component.scss
@@ -1,9 +1,4 @@
 .revalidation-form {
   min-width: 150px;
   max-width: 500px;
-  width: 100%;
-}
-
-.full-width {
-  width: 100%;
 }
--- a/src/app/dashboard/revalidation-form/revalidation-form.component.scss
+++ b/src/app/dashboard/revalidation-form/revalidation-form.component.scss
@@ -1,0 +1,9 @@
+.revalidation-form {
+  min-width: 150px;
+  max-width: 500px;
+  width: 100%;
+}
+
+.full-width {
+  width: 100%;
+}
--- a/src/app/dashboard/revalidation-history/revalidation-history.component.scss
+++ b/src/app/dashboard/revalidation-history/revalidation-history.component.scss
@@ -1,0 +1,57 @@
+@import "~nhsuk-frontend/packages/core/settings/colours";
+
+.spacer {
+  flex: 1 1 auto;
+}
+
+table {
+  width: 100%;
+}
+
+tr.detail-row {
+  height: 0;
+}
+
+tr.element-row:not(.expanded-row):hover {
+  background: $color_nhsuk-grey-5;
+}
+
+tr.element-row:not(.expanded-row):active {
+  background: $color_transparent_nhsuk-blue-50;
+}
+
+.element-row td {
+  border-bottom-width: 0;
+}
+
+.element-detail {
+  overflow: hidden;
+  display: flex;
+  flex-direction: column;
+}
+
+.outcome {
+  line-height: 24px;
+  display: flex;
+  vertical-align: middle;
+}
+
+.outcome-text {
+  margin-left: 8px;
+}
+
+.element-submission-status,
+.element-notes {
+  padding: 16px 0;
+}
+
+.element-description-attribution {
+  opacity: 0.5;
+}
+
+.button-row {
+  margin-top: 32px;
+  padding: 16px;
+  display: flex;
+  flex-direction: row;
+}
--- a/src/app/dashboard/revalidation-history/revalidation-history.component.scss
+++ b/src/app/dashboard/revalidation-history/revalidation-history.component.scss
@@ -4,10 +4,6 @@
   flex: 1 1 auto;
 }
 
-table {
-  width: 100%;
-}
-
 tr.detail-row {
   height: 0;
 }
--- a/src/app/dashboard/revalidation-history/revalidation-history.component.ts
+++ b/src/app/dashboard/revalidation-history/revalidation-history.component.ts
@@ -1,0 +1,112 @@
+import { Component, OnInit } from "@angular/core";
+import {
+  animate,
+  state,
+  style,
+  transition,
+  trigger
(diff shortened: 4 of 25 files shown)


ojemuyiwa:
> Please fix the commit prefixes(should be 'feat' instead of feature)

Will do

src/app/dashboard/revalidation-form/revalidation-form.component.scss
```diff
@@ -0,0 +1,9 @@
+.revalidation-form {
+  min-width: 150px;
+  max-width: 500px;
+  width: 100%;
+}
+
+.full-width {
```
omarCreativeDev: there is already a css class in utilities `.w-100 ` that does the same thing

src/app/dashboard/revalidation-form/revalidation-form.component.scss
```diff
@@ -0,0 +1,9 @@
+.revalidation-form {
+  min-width: 150px;
+  max-width: 500px;
+  width: 100%;
```
omarCreativeDev: this width could be removed and the `w-100` class can be leveraged instead

src/app/dashboard/revalidation-history/revalidation-history.component.scss
```diff
@@ -0,0 +1,57 @@
+@import "~nhsuk-frontend/packages/core/settings/colours";
+
+.spacer {
+  flex: 1 1 auto;
+}
+
+table {
+  width: 100%;
```
omarCreativeDev: this can also probs be removed if you use the shared class

src/app/dashboard/revalidation-history/revalidation-history.component.ts
```diff
@@ -0,0 +1,112 @@
...
+      console.log(`Dialog result: ${result}`);
+    });
+  }
+
+  toggleWatchTrainee(): void {
+    this.watchTrainee = !this.watchTrainee;
+  }
+}
+
+export interface PeriodicElement {
```
omarCreativeDev: i think it would be nice to move this interface inside /core inside an interfaces file so other parts of the app can consume

src/app/dashboard/dashboard.module.ts
```diff
@@ -6,9 +6,20 @@ import { DashboardRoutingModule } from "./dashboard-routing.module";
...
+import { RevalidationNotesComponent } from "./revalidation-notes/revalidation-notes.component";
+import { RevalidationFormComponent } from "./revalidation-form/revalidation-form.component";
 
 @NgModule({
-  declarations: [DashboardComponent, UnderNoticeComponent],
+  declarations: [
+    DashboardComponent,
+    UnderNoticeComponent,
+    TraineeSummaryComponent,
+    RevalidationHistoryComponent,
```
omarCreativeDev: i think as this details screen is not a child of dashboard it would make sense to create a separate module alongside a routing file

src/app/dashboard/revalidation-form/revalidation-form.component.scss
ojemuyiwa: done.

src/app/dashboard/revalidation-form/revalidation-form.component.scss
ojemuyiwa: done

src/app/dashboard/revalidation-history/revalidation-history.component.scss
ojemuyiwa: done

src/app/dashboard/revalidation-notes/revalidation-notes.component.ts
```diff
@@ -0,0 +1,39 @@
...
+})
+export class RevalidationNotesComponent implements OnInit {
+  messages: string[]; // mock
+  addEditNote: boolean;
+  constructor(
+    private bottomSheetRef: MatBottomSheetRef<RevalidationNotesComponent>
+  ) {}
+
+  ngOnInit(): void {
+    this.messages = [
```
omarCreativeDev: maybe one for a separate pr due to time and effort etc but this mock data should make its way to the state via ngxs

src/app/dashboard/revalidation-notes/revalidation-notes.component.ts
```diff
@@ -0,0 +1,39 @@
...
+import { MatBottomSheetRef } from "@angular/material/bottom-sheet";
+
+@Component({
+  selector: "app-revalidation-notes",
+  templateUrl: "./revalidation-notes.component.html",
+  styleUrls: ["./revalidation-notes.component.scss"]
+})
+export class RevalidationNotesComponent implements OnInit {
+  messages: string[]; // mock
+  addEditNote: boolean;
```
omarCreativeDev: same as below, this state should ideally live inside ngxs state management

src/app/dashboard/revalidation-history/revalidation-history.component.ts
ojemuyiwa: absolutely: unknown interfaces all around will refactor if similar can be returned from the back-end. They said they were going to give one massive data-source so i'm totally unsure of what to do for now

src/app/dashboard/revalidation-notes/revalidation-notes.component.scss
```diff
@@ -0,0 +1,32 @@
+.spacer {
```
omarCreativeDev: is this spacer class something generic or specific to this component? if its specific to this component probs be good to use BEM for naming the classes to avoid any overlap

src/app/dashboard/revalidation-notes/revalidation-notes.component.scss
```diff
@@ -0,0 +1,32 @@
...
+  flex: 1 1 auto;
+}
+
+.flex-container {
```
omarCreativeDev: same as above

src/app/dashboard/revalidation-history/revalidation-history.component.ts
```diff
@@ -0,0 +1,112 @@
...
+    "outcome",
+    "gmcSubDueDate",
+    "ActSubDate",
+    "submittedBy"
+  ];
+  expandedElement: PeriodicElement | null;
+  constructor(private bottomSheet: MatBottomSheet, public dialog: MatDialog) {}
+
+  ngOnInit(): void {
+    this.dataSource = [
```
omarCreativeDev: as before this is probs one for later but this mock data should really live in state management

src/app/dashboard/revalidation-history/revalidation-history.component.ts
```diff
@@ -0,0 +1,112 @@
...
+    this.watchTrainee = !this.watchTrainee;
+  }
+}
+
+export interface PeriodicElement {
+  recommendation: string;
+  episodeId: number;
+  outcome: string; // enum from GMC table
+  gmcSubDueDate: string;
+  ActSubDate: string;
```
omarCreativeDev: can this be camel case too?

src/app/dashboard/dashboard.module.ts
ojemuyiwa: it is actually

src/app/dashboard/revalidation-notes/revalidation-notes.component.ts
ojemuyiwa: yup agree

src/app/dashboard/revalidation-notes/revalidation-notes.component.ts
ojemuyiwa: agreed, and also table datasource etc. all hard coded for now

src/app/dashboard/revalidation-notes/revalidation-notes.component.scss
ojemuyiwa: no-overlap on encapsulated classes is possible. ideally what we'd want is a css structure review i.e. if we actually need individual component sass

src/app/dashboard/revalidation-history/revalidation-history.component.ts
ojemuyiwa: have no idea what casing back-end will return. Literally copied your ITrainee style

src/app/dashboard/revalidation-history/revalidation-history.component.ts
omarCreativeDev: i think we can expect these interfaces to change which is fine but i was more getting at maybe we should store them under /core but no worries something for later

src/app/dashboard/dashboard.module.ts
omarCreativeDev: hmmm the revalidation history route seems to be completely independant to the dasbhboard no? no worries happy to leave it as it is

PR #31, opened 2020-03-30
feat:add github actions and ecs support

paulhoang-hee:


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,82 @@
...
+          aws-region: eu-west-2
+
+      - name: Log in to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v1
+
+      - name: Docker Build, tag and push image to Amazon ECR
+        env:
+          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          ECR_REPOSITORY: tis-revalidation-v2
```
paulhoang-hee: if anyone knows how to lowercase and set an envar please let me know

Judge40: This for re-using the repo name for the ECR repo?

Judge40: If so, I'm happy to rename the repo. Did the same with `tis-trainee-details` for similar reasons.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,82 @@
...
+          IMAGE_TAG: ${{ github.sha }}
+        run: |
+          docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG .
+          docker push $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG
```
JosephKelly: I think it could be useful to have it also tagged as `latest`.  I'm thinking of how to simplify pulling images for running locally.

johnsimmons1980: good call

.github/workflows/deploy.yml
paulhoang-hee: yea, its so that we can create images with lowercase names, struggled with it in the other project so I just hardcoded it for now

.github/workflows/deploy.yml
paulhoang-hee: done

PR #32, opened 2020-04-02
feat: add sorting to dashboard

omarCreativeDev:
TISNEW-4120

- added sorting feature to trainees [need to investigate how to restrict sorting on columns and once i find out will raise a separate pr]
- ability to reflect the sort in the URL
- encapsulated trainees route in a separate module for better performance, also because the state management slice will be also self contained on the store
- updated and fixed unit and e2e tests that were effected due to the above changes
- created constants file so that default route can be reused instead of being hard coded
- general tidy up and refactor on method names etc, sorted npm scripts


src/app/trainees/trainees-routing.module.ts
```diff
@@ -0,0 +1,25 @@
...
+import { TraineeListComponent } from "./trainee-list/trainee-list.component";
+
+const routes: Routes = [
+  {
+    path: "trainees",
+    component: TraineesComponent,
+    data: { title: "Trainees" },
+    children: [
+      {
+        path: "column/:sortColumn/direction/:sortDirection",
```
ojemuyiwa, line 13: these have been made mandatory where I feel they are optional i.e. path could be "" with optional {sortcolumn, sortdirection}  which will look like "http://localhost:4200/trainees/;submissionDate;asc" I think we should discuss routing so the app is not disjointed. "http://localhost:4200/trainees/:id" should get trainee with id. now this isn't optional and "http://localhost:4200/trainees" should get all trainees, no filters etc.

omarCreativeDev: good point @ojemuyiwa - been doing some reading here https://angular-2-training-book.rangle.io/routing/query_params and it seems that query params are the way to go to make them optional as route param variables behave as essential to determining route

I have made the params optional via query strings which means now `http://localhost:4200/trainees` also gets all trainees with no filters or any other params.

pr updated

PR #39, opened 2020-04-07
feat: trainees list paginator

omarCreativeDev:
TISNEW-4186

- pagination feature for trainees list
- updated `IGetTraineesResponse` interface as per BE changes
- updated `getTrainees` FE mock endpoint
- fixed all outstanding warnings in unit testing around trainee details screen components
- removed redundant 2 tests from `AppComponent`

### DIFF
--- a/src/app/trainees/state/trainees.actions.ts
+++ b/src/app/trainees/state/trainees.actions.ts
@@ -26,3 +26,16 @@ export class SearchTrainees {
   static readonly type = "[Trainees] Search";
   constructor(public query: string) {}
 }
+
+export class PaginateTrainees {
+  static readonly type = "[Trainees] Page";
+  constructor(public pageIndex: number) {}
+}
+
+export class UpdateTraineesRoute {
+  static readonly type = "[Trainees] Update Route";
+}
+
+export class ResetTraineesPaginator {
+  static readonly type = "[Trainees] Reset Paginator";
+}
--- a/src/app/trainees/trainees.component.html
+++ b/src/app/trainees/trainees.component.html
@@ -5,3 +5,7 @@
 <div class="p-15">
   <router-outlet></router-outlet>
 </div>
+
+<div class="d-grid justify-content-center p-15">
+  <app-trainee-list-paginator class="pt-10 pb-10"></app-trainee-list-paginator>
+</div>
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -23,19 +23,4 @@ describe("AppComponent", () => {
     const app = fixture.componentInstance;
     expect(app).toBeTruthy();
   });
-
-  xit(`should have title as 'revalidation'`, () => {
-    const fixture = TestBed.createComponent(AppComponent);
-    const app = fixture.componentInstance;
-    expect(app.title).toEqual("revalidation");
-  });
-
-  xit("should render title", () => {
-    const fixture = TestBed.createComponent(AppComponent);
-    fixture.detectChanges();
-    const compiled = fixture.nativeElement;
-    expect(compiled.querySelector(".nhsuk-body").textContent).toContain(
-      "revalidation app is running!"
-    );
-  });
 });
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -4,6 +4,4 @@ import { Component } from "@angular/core";
   selector: "app-root",
   templateUrl: "./app.component.html"
 })
-export class AppComponent {
-  public title = "revalidation";
-}
+export class AppComponent {}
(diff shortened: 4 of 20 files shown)


src/app/trainees/state/trainees.actions.ts
```diff
@@ -26,3 +26,16 @@ export class SearchTrainees {
   static readonly type = "[Trainees] Search";
   constructor(public query: string) {}
 }
+
+export class PaginateTrainees {
+  static readonly type = "[Trainees] Page";
+  constructor(public pageIndex: number) {}
```
ojemuyiwa, line 32: just wondering: shouldn't there also be a, number of items per page? i.e. pageIndex and pageSize parameters? i.e. if we're defaulting to pageSize = 20

src/app/trainees/trainees.component.html
```diff
@@ -5,3 +5,7 @@
 <div class="p-15">
   <router-outlet></router-outlet>
 </div>
+
```
ojemuyiwa, line 8: Tying component instances with routing may have an adverse effect with events. e.g. ngOnInit, etc will fire at separate instances out of sync if event based data. Just thinking aloud if clearFilters, pageChanges etc affect routerNagivationEnd, then data will be out of sync. If components are all loaded in tranee-list @inputs and outputs may keep them in sync? Maybe? trainee lists may keep subscribing to parameter changes (this.route.queryParamMap.subscribe || this.route.queryParams.subscribe) and fire actions appropriately? I will download this branch and see events sequences with sorting and page changing to see ... once again just thinking aloud

src/app/trainees/state/trainees.actions.ts
omarCreativeDev: the `pageSize` of 20 is passed in via an input on the `<mat-paginator>`

ojemuyiwa: says pageSize=1 on <mat-paginator>

ojemuyiwa: And this is not passed to Uzairs code to return 20 items? Thats what I was alluding to

src/app/trainees/trainees.component.html
omarCreativeDev: i dont think i quite understand. the route change is basically done to gracefully update the url which doesn't reinstantiate the component or fire ngOnInit either. 

the queryParams check via url snapshot is to cater for the scenario if `/trainees` route is viewed without any parameters it lets the store know how to update appropriately and subsequently make the api call with the correct default parameters

src/app/trainees/state/trainees.actions.ts
omarCreativeDev: ah yea my bad, good spot. just updated the pr with 20

src/app/trainees/trainees.component.html
ojemuyiwa: yeah i just checked, we're firing quite a bit in different events.

Screenshot 2020-04-07 at 11 35 57

ojemuyiwa: i'll pull-down and check again. I suspect we're talking about different things here

ojemuyiwa: > i dont think i quite understand. the route change is basically done to gracefully update the url which doesn't reinstantiate the component or fire ngOnInit either.
> 
> the queryParams check via url snapshot is to cater for the scenario if `/trainees` route is viewed without any parameters it lets the store know how to update appropriately and subsequently make the api call with the correct default parameters

In response: ngOnint always fires on route change if your component loads in router-outlet

omarCreativeDev: the BE work for pagination hasn't completed yet so the api response is sending back thousands of records and the ui therefore has a super long almost never ending scroll. i would imagine the paginator is out of view but not to worry as I've got another task to integrate the pagination with the BE once uzair is done. 

ngOnInit doesn't fire when navigating to the same route which is what is happening on this page

ojemuyiwa: Screenshot 2020-04-07 at 16 26 27

On Initial load, OnInit in pagination fires before OnInit in trainee-list.

This means is I hard code variables in url and hard refresh pagination control sends action first 
`if (params.pageIndex) {
      this.store.dispatch(new PaginateTrainees(params.pageIndex));
    } else {
      this.store.dispatch(new ResetTraineesPaginator());
    }`
then trainee-list sends its own OnInit
`if (params.active && params.direction) {
      this.store.dispatch(new SortTrainees(params.active, params.direction));
    } else {
      this.store.dispatch(new ResetTraineesSort());
    }

this.store.dispatch(new GetTrainees());
`

ojemuyiwa: Screenshot 2020-04-07 at 16 27 51

PR #41, opened 2020-04-08
illustrate ngOnInit doesnt fire on same route change - please do not merge

omarCreativeDev:
please do not merge, pr to illustrate ngOnInit doesnt fire on same route change

- ngxs console logs disabled and coloured console logs added to paginator and trainee list components ngOnInit
- notice when navigating to same route via clicking on sorting, paginator or clear all elements no ngOnInit's fire and therefore no new console logs

### DIFF
--- a/src/app/shared/shared.module.ts
+++ b/src/app/shared/shared.module.ts
@@ -13,8 +13,7 @@ import { PageNotFoundComponent } from "./page-not-found/page-not-found.component
     CommonModule,
     ReactiveFormsModule,
     RouterModule,
-    NgxsReduxDevtoolsPluginModule.forRoot({ disabled: environment.production }),
-    NgxsLoggerPluginModule.forRoot({ disabled: environment.production })
+    NgxsReduxDevtoolsPluginModule.forRoot({ disabled: environment.production })
   ],
   exports: [CommonModule, ReactiveFormsModule]
 })
--- a/src/app/trainees/trainee-list-paginator/trainee-list-paginator.component.ts
+++ b/src/app/trainees/trainee-list-paginator/trainee-list-paginator.component.ts
@@ -27,6 +27,10 @@ export class TraineeListPaginatorComponent implements OnInit {
    * And update store accordingly
    */
   ngOnInit(): void {
+    console.log(
+      "%c TraineeListPaginatorComponent ngOnInit fired ",
+      "background: orange; color: black"
+    );
     const params: Params = this.route.snapshot.queryParams;
 
     if (params.pageIndex) {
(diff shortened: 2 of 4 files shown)


ojemuyiwa:
> please do not merge, pr to illustrate ngOnInit doesnt fire on same route change
> 
> * ngxs console logs disabled and coloured console logs added to paginator and trainee list components
> * notice when navigating to same route via clicking on sorting, paginator or clear all elements no ngOnInit's fire and therefore no new console logs

Agreed. In the effort to prove this right you are 1000% correct. route parameters both required and non-required reload components, queryParams don't. "This is actually a bug as URI specifications you may well know were made for this, hence we all capitalize on it in cache-bursting" I digress. This solution may work in some scenarios, but If the purpose of adding all these parameters to URI is so users can share links. You Add pageNumber=5 etc and copy and paste in another link: Then watch the pageIndex which should be pageNumber fire onInit before the sort fire OnInit and in console see if we have 1 unified querycombined call to the back-end. 
A pull request is just for sanity second pair of eyes check, when I get let x:string = ""; please remove string as it's implied.... I don't go off on one. This is just an opinion in linting quite frankly 2*6 = 1*12. 
All I pointed out were potential pit-falls e.g. Navigation within store has to be run in ngZone, once you see this its a pointer navigation shouldn't be handled there either because you could be stuck in an infinite loop. But to keep the peace I've pushed the PR through, bugs and all.. we'll see how it plays out. Trainee get ideally should have constructor with ALL parameters and some defaults set.  because every call is a GET with filter,pageNumber,order,coulumnSort etc...... on the same api url. Reset sets back to defaults. These are potential (Key word, potential) issues I see with the logic.

omarCreativeDev:
if the url is copied, pasted & manually navigated e.g `trainees?active=submissionDate&direction=desc&pageIndex=5` the components on the page init once and then update the store with values from the query params and then BE api is called with these same param values.

in the other scenario where there are no query params i.e `/trainees` then OnInit fires once and sets default values in store [as theres nothing in query params] and then makes BE api call with these default values.

omarCreativeDev:
good point about navigation within store having to run in ngZone. will take a look at this at some point today and raise a pr to address

ojemuyiwa:
> if the url is copied, pasted & manually navigated e.g `trainees?active=submissionDate&direction=desc&pageIndex=5` the components on the page init once and then update the store with values from the query params and then BE api is called with these same param values.
> 
> in the other scenario where there are no query params i.e `/trainees` then OnInit fires once and sets default values in store [as theres nothing in query params] and then makes BE api call with these default values.

cool stuff

PR #43, opened 2020-04-08
chore: store should route with ngzone

omarCreativeDev:
TISNEW-4194

- ensured store changes route with ngZone [fixes console warning]
- updated unit tests 
- ensured route change action is dispatched after GetTrainees has resolved
- dispatched actions in trainee-list, paginator and reset-trainee-list components individually as opposed to multiple actions in an array. this improves the stores development mode console logging and makes it much nicer to debug;

Screenshot 2020-04-08 at 18 37 53

### DIFF
--- a/src/app/trainees/state/trainees.state.ts
+++ b/src/app/trainees/state/trainees.state.ts
@@ -1,10 +1,10 @@
 import { HttpParams } from "@angular/common/http";
-import { Injectable } from "@angular/core";
+import { Injectable, NgZone } from "@angular/core";
 import { Sort } from "@angular/material/sort";
 import { Router } from "@angular/router";
 import { State, Action, StateContext, Selector } from "@ngxs/store";
 import { tap } from "rxjs/operators";
-import { DEFAULT_ROUTE_SORT } from "../../core/trainee/constants";
+import { DEFAULT_SORT } from "../../core/trainee/constants";
 import { ITrainee } from "../../core/trainee/trainee.interfaces";
 import { TraineeService } from "../../core/trainee/trainee.service";
 import {
@@ -39,8 +39,11 @@ export class TraineesStateModel {
 })
 @Injectable()
 export class TraineesState {
-  public defaultSort: Sort = DEFAULT_ROUTE_SORT;
-  constructor(private traineeService: TraineeService, private router: Router) {}
+  constructor(
+    private traineeService: TraineeService,
+    private router: Router,
+    private ngZone: NgZone
+  ) {}
 
   @Selector()
   public static trainees(state: TraineesStateModel) {
@@ -113,7 +116,7 @@ export class TraineesState {
     const state = ctx.getState();
     return ctx.setState({
       ...state,
-      sort: this.defaultSort
+      sort: DEFAULT_SORT
     });
   }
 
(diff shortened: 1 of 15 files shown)


src/app/trainees/state/trainees.state.ts
```diff
@@ -141,12 +145,14 @@ export class TraineesState {
...
   updateTraineesRoute(ctx: StateContext<TraineesStateModel>) {
     const state = ctx.getState();
-    return this.router.navigate(["/trainees"], {
-      queryParams: {
-        active: state.sort.active,
-        direction: state.sort.direction,
-        pageIndex: state.pageIndex
-      }
-    });
+    return this.ngZone.run(() =>
```
ojemuyiwa, line 148: hard coded navigation in store seems off

omarCreativeDev: be hardcoded navigation did you mean the route string `/trainees`? good spot if this is what you meant, just replaced with existing variable `DEFAULT_ROUTE`

ojemuyiwa: Just a thought what happens when default route is totally taken out?

router.navigate(["./"], queryParams, relativeTo: ActivatedRoute
or router.navigate({queryParams, relativeTo: ActivatedRoute}

will always goto the same route and change whatever was passed through, was thinking of using same approach for under-notice route hence previous conversations about routing strategies.

ojemuyiwa: Just a thought: 

what happens when default route is taken out?

router.navigate(["./"], {queryParams, relativeTo: AtivatedRoute})
or
router.navigate({queryParams, relativeTo: AtivatedRoute})

I ask because route independent e.g. /trainees/under-notice with same component was discussed but may not work with all the hard coded default route code. 

Let me know your thoughts.

omarCreativeDev: how do you mean when you say default route taken out? any attempts to navigate to a route that doesn't exist does / should redirect to a page not found.

so far the /trainees route has been built on the basis that sub routes are not needed as Ify clarified the data table and ui remains identical whether filtering via under notice or all doctors. essentially under notice and all doctors are just a filter thats appended to the api call which as a result updates the data table

ojemuyiwa: What I mean is its not route centric. From dashboard, to trainees to maybe doctors tomorrow it is decoupled. If default route is "/" in application for instance we would not need to have DEFAULT_ROUTE all over the application. It's a hack. If we put component on another route state management will be redirecting to DEFAULT_ROUTE its too tightly coupled hence I advised not to have routing within state-management. Also the reset functionality will reset underNotice=true in url parameters if we decide to go this way. Hence the questions.

ojemuyiwa: I just think a reset parameters on activated route may be a better solution. Let me know your thoughts and if you'd like to proceed with default-route

ojemuyiwa: Final answer? Go with DEFAULT_ROUTE? yay or nay? lemme know i'll approve and merge

omarCreativeDev: no worries, I have gone ahead and removed the `DEFAULT_ROUTE` variable altogether. note that whilst we can now navigate to `/` [e.g from header or from page not found route] which as a fallback redirects to `/trainees` with default params as per app-routing.module.

in order to retain or update the query params in the url routing should be performed toto  `/trainees` alongside the optional params.

pr updated

PR #45, opened 2020-04-11
Tisnew 4154 under notice tab review do not merge

ojemuyiwa:
For Omar: FE refactor. Have a look as per multiple discussions. External Modules running outside NGZone stay outside NGZone. Reduced boiler-plate code and increased code-coverage which could be much more improved if review is accepted.


ojemuyiwa:
> i haven't had a chance to look at this pr in detail due to its size but here are my initial observations;
> 
> * i noticed you have `please do not merge` in the title, is this meant to be a draft pr?
> * there's lots of conflicting files so please first rebase from master
> * not sure why you have added nested routes back in for `/trainees`? we discussed this previously and both agreed that nested routes are not needed. you asked me to remove the `<router-outlet>` as i did in this pr #44
> * adding a resolver to the trainees route will essentially block the ui which when we discussed previously with the wider team the other feds and myself were not keen on this idea. better to asynchronously load in and update the bits of the ui when and where needed alongside spinner/s rather then blocking the whole page from loading.
> * whats happened to the state actions? Looks like you've deleted all the actions except `GetTrainees`. I dont think its a good idea to cram so many events in to 1. We should leverage the advantages of the redux state management pattern that NGXS offers.

This is just for discussion. It has filtering, pagination, search, routing which allows default params for under-notice which we had previously discussed, no hard coded DEFAULT_ROUTE, global spinner, localhost reference to local build, less boiler-plate code on state management.

As stated its a discussion we should have as theres no hard-coding reroute /trainees etc as I cannot think of a way without hacking the planet to deliver under-notice

omarCreativeDev:
No worries I understand. The DEFAULT_ROUTE variable holds the string value of the `/trainees` route which has been re used in multiple places in the app. I thought a variable to be reused is better than the same string hard coded in multiple places. Which state management boiler plate code are you referring to? One improvement I have started to make in my recent pr https://github.com/Health-Education-England/TIS-REVALIDATION-V2/pull/46 is to leverage patchState over setState which feels a bit more cleaner.

I did pull the branch down and ran the app and also noticed that you've merged trainees and dashboard modules back in to one? The reason I create a seperate module for trainees module is for performance reasons. As because for example the details page doesn't consume or need the components that are displayed on the trainees page and vice versa so makes sense to have 2 lightweight js bundles otherwise each route is loading in code that it doesn't need. Plus it means we dont end up with one massive module that becomes harder to split down the line if and when the app grows.

Under notice & all doctors could require there own action/s - something like `UnderNoticeTraineesFilter` & `AllDoctorsTraineesFilter` or perhaps 1 action as `TraineesFilter` with a boolean payload as BE essentially expects `underNotice=<true|false>` as a query param? The store would then react to this event and patch the state. Then you'd have to append this param in the `GetTrainees` action listener in a similar fashion to how the other params are.

So basically clicking on one of the filter tabs would dispatch the filter action, `GetTrainees` action and then once resolved `UpdateTraineesRoute` action [to gracefully update the query params in the url]. This will essentially result in an api call being made with the required params, loading spinner in the meantime [as its bound to the async api call], and then the data in the table refreshed, and then finally the query params are updated in the url.

PR #46, opened 2020-04-13
feat: trainee list search

omarCreativeDev:
TISNEW-4054

- added new search feature for trainees
- includes form validation
- includes unit tests
- ensured `searchQuery` is appended to api call
- ensured reset trainees list `Clear All` also resets search
- moved pagination query params check within `trainee-list` to ensure all query params are checked in same component upon ngOnInit life cycle

### DIFF
--- a/src/app/trainees/state/trainees.actions.ts
+++ b/src/app/trainees/state/trainees.actions.ts
@@ -24,7 +24,11 @@ export class ClearTraineesFilter {
 
 export class SearchTrainees {
   static readonly type = "[Trainees] Search";
-  constructor(public query: string) {}
+  constructor(public searchQuery: string) {}
+}
+
+export class ClearTraineesSearch {
+  static readonly type = "[Trainees] Clear Search";
 }
 
 export class PaginateTrainees {
--- a/src/app/trainees/trainee-search/trainee-search.component.html
+++ b/src/app/trainees/trainee-search/trainee-search.component.html
@@ -1,0 +1,14 @@
+<form (ngSubmit)="checkForm()" [formGroup]="form" autocomplete="off">
+  <mat-form-field>
+    <input
+      matInput
+      placeholder="Enter name / GMC no"
+      [formControlName]="'searchQuery'"
+      [value]="searchQuery$ | async"
+    />
+    <mat-error *ngIf="form.invalid">No search terms entered</mat-error>
+  </mat-form-field>
+  <button mat-button type="submit">
+    <mat-icon>search</mat-icon>
+  </button>
+</form>
--- a/src/app/trainees/trainee-search/trainee-search.component.ts
+++ b/src/app/trainees/trainee-search/trainee-search.component.ts
@@ -1,0 +1,58 @@
+import { Component, OnInit } from "@angular/core";
+import { FormBuilder, FormGroup, Validators } from "@angular/forms";
+import { ActivatedRoute, Params } from "@angular/router";
+import { Select, Store } from "@ngxs/store";
+import { Observable } from "rxjs";
+import { take } from "rxjs/operators";
+import {
+  GetTrainees,
+  SearchTrainees,
+  UpdateTraineesRoute
+} from "../state/trainees.actions";
+import { TraineesState } from "../state/trainees.state";
+
+@Component({
+  selector: "app-trainee-search",
+  templateUrl: "./trainee-search.component.html"
+})
+export class TraineeSearchComponent implements OnInit {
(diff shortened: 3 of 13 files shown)


src/app/trainees/trainee-search/trainee-search.component.html
```diff
@@ -0,0 +1,14 @@
+<form (ngSubmit)="checkForm()" [formGroup]="form" autocomplete="off">
+  <mat-form-field>
+    <input
+      matInput
+      placeholder="Enter name / GMC no"
+      [formControlName]="'searchQuery'"
+      [value]="searchQuery$ | async"
+    />
+    <mat-error *ngIf="form.invalid">Minimum 3 characters needed</mat-error>
```
ojemuyiwa, line 9: unsure of min chars validation needed. Chinese doctor named QI or YI cannot be searched for? know the requirement said don't auto-search until 3 characters but this is just my thoughts

src/app/trainees/trainee-search/trainee-search.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    private route: ActivatedRoute
+  ) {}
+
+  ngOnInit() {
+    this.setupForm();
+  }
+
+  public setupForm(): void {
+    this.form = this.formBuilder.group({
+      searchQuery: [this.params.searchQuery, [Validators.minLength(3)]]
```
ojemuyiwa, line 35: discuss validator if its really needed.

src/app/trainees/trainee-search/trainee-search.component.html
ojemuyiwa: will approve on your request. this is just food for thought. lemme know if its a go?

omarCreativeDev: good point, i'll reach out and ask Ify on the jira ticket

src/app/trainees/trainee-search/trainee-search.component.ts
omarCreativeDev: ify has updated the jira ticket so minLength is now set at 2 characters. pr updated

src/app/trainees/state/trainees.actions.ts
```diff
@@ -24,7 +24,11 @@ export class ClearTraineesFilter {
 
 export class SearchTrainees {
   static readonly type = "[Trainees] Search";
-  constructor(public query: string) {}
+  constructor(public searchQuery: string) {}
```
babul3012, line 27: What is the need for this arguments  to be public?

omarCreativeDev: @babul3012 the args of the actions need to be public so that the `TraineesState` class can access them

src/app/trainees/trainee-search/trainee-search.component.html
ojemuyiwa: Let me know when you're ready with this as well .... As for form control default auto search I'd suggest something along the lines of 
`searchQuery.valueChanges
      .pipe(debounceTime(1000))
      .subscribe((searchTerm: any) => {
        // do magic to put search term in url
      });`
I await till Ify's revision and your response

omarCreativeDev: yea i was thinking about using `valueChanges` on the form control as well but felt subscribing to every key stroke [even with a debounce] was not needed. at the moment if the mouse focus is on the form the enter key or submit button both run the `checkForm()` method which basically trims the search string and then if form is valid allows to proceed with dispatching the action and then hitting the BE.

ify updated the ticket earlier and changed the min character length to 2 - pr is updated to reflect

src/app/trainees/trainee-search/trainee-search.component.ts
```diff
@@ -0,0 +1,58 @@
...
+    private route: ActivatedRoute
+  ) {}
+
+  ngOnInit() {
+    this.setupForm();
+  }
+
+  public setupForm(): void {
+    this.form = this.formBuilder.group({
+      searchQuery: [this.params.searchQuery, [Validators.minLength(2)]]
```
ojemuyiwa, line 35: Screenshot 2020-04-14 at 16 10 52

please confirm and I'll approve.

ojemuyiwa: Doubt this meant validation but rather Observable debounce and check value length >= 2.

I may be off, but confirm and I'll approve.

omarCreativeDev: ah good spot, will clarify with Ify

PR #47, opened 2020-04-13
fix: local dev environment fixes

ojemuyiwa:
localhost replace dev local home IP
scss reduced into utilities
environment variables refactored for GA, hotjar and removed redundant host
desktop menu linked to dev reval app and css z-index fix

### DIFF
--- a/src/app/shared/main-navigation/mat-main-nav/menu-items.const.ts
+++ b/src/app/shared/main-navigation/mat-main-nav/menu-items.const.ts
@@ -83,31 +83,19 @@ export const menuItems: IMenuItem[] = JSON.parse(
     {
         "name": "Revalidation",
         "type": 0,
-        "route": "admin/manage-connections",
+        "route": "/trainees",
         "description": "",
         "menuItems": [
             {
                 "type": 0,
-                "route": "admin/manage-connections",
-                "name": "Manage connections",
+                "route": "/trainees",
+                "name": "All Trainees",
                 "description": ""
             },
             {
                 "type": 0,
-                "route": "concerns/",
-                "name": "Concerns log",
-                "description": ""
-            },
-            {
-                "type": 0,
-                "route": "admin/manage-connections/create",
-                "name": "Connection discrepancies",
-                "description": ""
-            },
-            {
-                "type": 0,
-                "route": "revalidation/",
-                "name": "Provide recommendation",
+                "route": "/trainees/under-notice",
+                "name": "Trainees under-notice",
                 "description": ""
             }
         ]
--- a/src/app/shared/main-navigation/mat-main-nav/menu-items.const.ts
+++ b/src/app/shared/main-navigation/mat-main-nav/menu-items.const.ts
@@ -88,14 +88,8 @@ export const menuItems: IMenuItem[] = JSON.parse(
         "menuItems": [
             {
                 "type": 0,
-                "route": "/trainees",
+                "route": "/",
                 "name": "All Trainees",
-                "description": ""
-            },
-            {
-                "type": 0,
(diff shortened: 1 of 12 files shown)


src/app/shared/main-navigation/mat-main-nav/menu-items.const.ts
```diff
@@ -83,31 +83,19 @@ export const menuItems: IMenuItem[] = JSON.parse(
...
-                "type": 0,
-                "route": "admin/manage-connections/create",
-                "name": "Connection discrepancies",
-                "description": ""
-            },
-            {
-                "type": 0,
-                "route": "revalidation/",
-                "name": "Provide recommendation",
+                "route": "/trainees/under-notice",
```
omarCreativeDev, line 97: `/trainees/under-notice` doesn't exist as a route. im guessing you meant something like `/trainees?<filterParamNameHere>=under-notice`?

src/environments/environment.interface.ts
```diff
@@ -1,7 +1,8 @@
 export abstract class IEnvironment {
   abstract readonly production: boolean; // build mode
   abstract readonly siteIds: string[]; // Google analytics site id's
-  abstract readonly host: string; // api proxy host
```
omarCreativeDev, line 4: removing the host seems like the right thing to do. can i ask what was your thinking behind this?

src/environments/environment.local.ts
```diff
@@ -4,14 +4,15 @@
 import { IEnvironment } from "./environment.interface";
 export const environment: IEnvironment = {
   production: false,
-  host: `api/`,
-  siteIds: [],
+  siteIds: ["UA-40570867-6"],
+  hotJarId: 1662399,
```
omarCreativeDev, line 8: should our local machines be using the same `hotJarId` and `hotJarSv` as other environments? im guessing we dont have unique values for each environment yet as our environments are setup yet?

src/app/shared/main-navigation/mat-main-nav/menu-items.const.ts
ojemuyiwa: Yeah thought it would after our many routing strategy discussions. Previous routes in diff didnt exist either

src/environments/environment.interface.ts
ojemuyiwa: there's no host. was used for paths e.g. api/

src/environments/environment.local.ts
ojemuyiwa: no dev hotjar or analytics variables. only protected with env.production from loading. as in prev diffs. moved to variable rather than hard coded

src/app/shared/main-navigation/mat-main-nav/menu-items.const.ts
omarCreativeDev: alternatively this value could just be `/` which will redirect to `/trainees` without the optional params?

src/environments/environment.interface.ts
omarCreativeDev: fair enough. much cleaner this way

src/environments/environment.local.ts
omarCreativeDev: fair enough, better to have variable for sure. i imagine we will get the environment specific values once the actual environments are ready

src/app/shared/main-navigation/mat-main-nav/menu-items.const.ts
ojemuyiwa: ok, i'll take the navigation item out as till route exists

PR #48, opened 2020-04-15
chore: move route change out from store

omarCreativeDev:
moved `/trainees` route change out of store as recommended by Femi to a service in order to ensure state management focuses on state

- removed `UpdateTraineesRoute` action as no longer needed
- small code refactor including tests
- added tests for `TraineeService`

### DIFF
--- a/src/app/core/trainee/trainee.service.spec.ts
+++ b/src/app/core/trainee/trainee.service.spec.ts
@@ -1,0 +1,64 @@
+import {
+  HttpClientTestingModule,
+  HttpTestingController
+} from "@angular/common/http/testing";
+import { TestBed } from "@angular/core/testing";
+import { Router } from "@angular/router";
+import { RouterTestingModule } from "@angular/router/testing";
+import { environment } from "@environment";
+import { NgxsModule, Store } from "@ngxs/store";
+import {
+  TraineesState,
+  TraineesStateModel
+} from "../../trainees/state/trainees.state";
+import { TraineeService } from "./trainee.service";
+
+describe("TraineeService", () => {
+  let service: TraineeService;
+  let http: HttpTestingController;
+  let router: Router;
+  let store: Store;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      imports: [
+        HttpClientTestingModule,
+        RouterTestingModule,
+        NgxsModule.forRoot([TraineesState])
+      ],
+      providers: [TraineeService]
+    });
+    service = TestBed.inject(TraineeService);
+    http = TestBed.inject(HttpTestingController);
+    router = TestBed.inject(Router);
+    store = TestBed.inject(Store);
+  });
+
+  it("should be created", () => {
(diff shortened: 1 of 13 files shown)


src/app/core/trainee/trainee.service.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    expect(service).toBeTruthy();
+  });
+
+  it("`getTrainees()` should call api", () => {
+    service.getTrainees().subscribe();
+
+    const mockHttp = http.expectOne(`${environment.appUrls.getTrainees}`);
+    expect(mockHttp.request.method).toBe("GET");
+
+    http.verify();
```
ojemuyiwa, line 47: metal note: mock variable if going to be used often could be verified automatically after e.g.

`afterEach(() => {
    httpMock.verify();
  });`

src/app/core/trainee/trainee.service.ts
```diff
@@ -1,19 +1,38 @@
...
+  ) {}
 
   public getTrainees(params?: HttpParams): Observable<IGetTraineesResponse> {
     return this.http.get<IGetTraineesResponse>(
       `${environment.appUrls.getTrainees}`,
       { params }
     );
   }
+
+  public updateTraineesRoute(): Promise<boolean> {
```
ojemuyiwa, line 27: fantastic solution

src/app/trainees/trainee-list-paginator/trainee-list-paginator.component.spec.ts
```diff
@@ -4,20 +4,28 @@ import { async, ComponentFixture, TestBed } from "@angular/core/testing";
...
-  UpdateTraineesRoute
-} from "../state/trainees.actions";
+import { Observable, of } from "rxjs";
+import { TraineeService } from "../../core/trainee/trainee.service";
+import { GetTrainees, PaginateTrainees } from "../state/trainees.actions";
 import { TraineesState } from "../state/trainees.state";
 
 import { TraineeListPaginatorComponent } from "./trainee-list-paginator.component";
 
+export class MockTraineeService {
```
ojemuyiwa, line 14: no need for export unless re-used in different tests? Could be in separate file if its going to be re-used and imported?

src/app/core/trainee/trainee.service.spec.ts
omarCreativeDev: ah yes very good point, will bear this in mind if and when there's multiple api calls to be tested in a service

src/app/trainees/trainee-list-paginator/trainee-list-paginator.component.spec.ts
omarCreativeDev: ah yes very good catch, cheers. using the export keyword has become bit of a habit oops. i think best to keep mocks within the spec itself otherwise the tests can become brittle by creating a dependancy.

pr updated

PR #52, opened 2020-04-20
feat: trainees list filter

omarCreativeDev:
TISNEW-4154 - implemented trainees list filtering

- added new `AllDoctorsFilter` and `UnderNoticeFilter` actions, selectors and related events
- added new `TraineeFiltersComponent` alongside some minor ui/scss tweaks
- extracted generation of http params in to service so code can be re used and easier unit testing
- tweaked and added more unit tests

### DIFF
--- a/src/app/core/trainee/trainee.service.ts
+++ b/src/app/core/trainee/trainee.service.ts
@@ -24,6 +24,22 @@ export class TraineeService {
     );
   }
 
+  public generateParams(): HttpParams {
+    const snapshot: TraineesStateModel = this.store.snapshot().trainees;
+
+    let params: HttpParams = new HttpParams()
+      .set("sortColumn", snapshot.sort.active)
+      .set("sortOrder", snapshot.sort.direction)
+      .set("pageNumber", snapshot.pageIndex.toString())
+      .set("underNotice", snapshot.underNotice.toString());
+
+    if (snapshot.searchQuery) {
+      params = params.append("searchQuery", snapshot.searchQuery);
+    }
+
+    return params;
+  }
+
   public updateTraineesRoute(): Promise<boolean> {
     const snapshot: TraineesStateModel = this.store.snapshot().trainees;
 
@@ -32,6 +48,7 @@ export class TraineeService {
         active: snapshot.sort.active,
         direction: snapshot.sort.direction,
         pageIndex: snapshot.pageIndex,
+        underNotice: snapshot.underNotice,
         ...(snapshot.searchQuery && { searchQuery: snapshot.searchQuery })
       }
     });
--- a/src/app/trainees/reset-trainee-list/reset-trainee-list.component.html
+++ b/src/app/trainees/reset-trainee-list/reset-trainee-list.component.html
@@ -1,1 +1,3 @@
-<span class="cursor-pointer" (click)="resetTraineeList()">Clear all</span>
+<button mat-button (click)="resetTraineeList()" aria-label="Clear all">
+  Clear all
+</button>
--- a/src/app/trainees/trainee-filters/trainee-filters.component.html
+++ b/src/app/trainees/trainee-filters/trainee-filters.component.html
@@ -1,0 +1,25 @@
+<ng-container
+  *ngIf="{
+    countTotal: countTotal$ | async,
+    countUnderNotice: countUnderNotice$ | async,
+    underNotice: underNotice$ | async
+  } as data"
+>
+  <button
+    mat-raised-button
+    [class.mat-primary]="!data.underNotice"
+    class="trainee-filters__btn"
+    aria-label="Filter all doctors"
+    (click)="filterByAllDoctors()"
+  >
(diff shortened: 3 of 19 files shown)


src/app/core/trainee/trainee.service.ts
```diff
@@ -24,6 +24,22 @@ export class TraineeService {
...
   }
 
+  public generateParams(): HttpParams {
+    const snapshot: TraineesStateModel = this.store.snapshot().trainees;
+
+    let params: HttpParams = new HttpParams()
+      .set("sortColumn", snapshot.sort.active)
+      .set("sortOrder", snapshot.sort.direction)
+      .set("pageNumber", snapshot.pageIndex.toString())
+      .set("underNotice", snapshot.underNotice.toString());
```
ojemuyiwa, line 34: wrapping this like search query seems like the right approach to adding under notice filter or not

src/app/trainees/trainee-filters/trainee-filters.component.html
```diff
@@ -0,0 +1,25 @@
+<ng-container
+  *ngIf="{
+    countTotal: countTotal$ | async,
+    countUnderNotice: countUnderNotice$ | async,
+    underNotice: underNotice$ | async
+  } as data"
+>
+  <button
```
ojemuyiwa, line 8: which button is active and which is in-active state wise? mat-tabs or routeActiveLink may be a better way to go for uniformity, but whichever way we go lets have an active and inactive state so users know what filter is active.

src/app/core/trainee/trainee.service.ts
omarCreativeDev: cheers for flagging, just double checked with BE on team reval slack and underNotice=false is the way to return all doctors

src/scss/utilities/_spacing-helpers.scss
```diff
@@ -19,7 +19,7 @@ $space-prefixes: (
   @each $attr-short, $attr-long in $space-prefixes {
     @each $value in $space-values {
       .#{$attr-short}-#{$value} {
-        #{$attr-long}: #{$value}#{"px"};
+        #{$attr-long}: #{$value}#{"px"} !important;
```
ojemuyiwa, line 22: important is not usually recommended maybe we can increase css priority if something needs overriding?

src/app/trainees/trainee-filters/trainee-filters.component.html
omarCreativeDev: ah yea good point but am setting the active state via the `mat-primary` class which is bound to `underNotice` true|false from store

src/scss/utilities/_spacing-helpers.scss
omarCreativeDev: yea good point. the !important is very ugly but had to override some out of the box angular material css. will investigate a heavier css selector perhaps as that would be cleaner

src/app/core/trainee/trainee.service.ts
ojemuyiwa: my guess with working with filters underNotice=false will return a sub-set of doctors where underNotice=false i.e. Not All Doctors

omarCreativeDev: as far as I know from uzair `underNotice=false` returns all doctors. is that correct @uzairbaig?

src/app/trainees/reset-trainee-list/reset-trainee-list.component.html
```diff
@@ -1 +1,3 @@
-<span class="cursor-pointer" (click)="resetTraineeList()">Clear all</span>
+<button mat-button (click)="resetTraineeList()" aria-label="Clear all">
+  Clear all
```
ojemuyiwa, line 2: Seems the reset may need to be revised if underNotice=false isn't default i.e. only if

src/app/trainees/reset-trainee-list/reset-trainee-list.component.ts
```diff
@@ -24,9 +24,9 @@ export class ResetTraineeListComponent {
   constructor(private store: Store, private traineeService: TraineeService) {}
```
ojemuyiwa, line 24: The clear / reset functionality seems not to set to default which I thought was `sortColumn=submissionDate&sortOrder=desc&pageNumber=0`

ojemuyiwa, line 24: The clear / reset functionality seems not to set to default which I thought was sortColumn=submissionDate&sortOrder=desc&pageNumber=0

src/scss/utilities/_spacing-helpers.scss
ojemuyiwa: let me know if you want to override in component itself or revise this later as in the grand scheme of things is not a show stopper.

omarCreativeDev: !important removed and pr updated

PR #54, opened 2020-04-20
reafactor: improve trainee list filtering

omarCreativeDev:
improve trainee list filtering for clarity

- rename `underNotice` state slice & query param to `filter` 
- created `TraineesFilterType`
- update `filter` query param boolean values to adhere to `TraineesFilterType`  
- updated unit tests


omarCreativeDev:
yup it works fine. at the moment the trainees list either displays all doctors or under notice doctors, there is no other variant. therefore the `underNotice` query parameter as true|false is always sent when the api is called in order to show all|under notice doctors.

yea i think there could be more filters coming in the future hence i renamed `underNotice` state slice to `filter` to going forward tweaks can be easily made

babul3012:
'chore' prefix in commit message if for  - updating build scripts/tasks, no production code changes.

src/app/core/trainee/trainee.service.ts
```diff
@@ -31,7 +31,7 @@ export class TraineeService {
       .set("sortColumn", snapshot.sort.active)
       .set("sortOrder", snapshot.sort.direction)
       .set("pageNumber", snapshot.pageIndex.toString())
-      .set("underNotice", snapshot.underNotice.toString());
+      .set("underNotice", snapshot.filter === "underNotice" ? "true" : "false");
```
babul3012, line 34: Can't we set boolean flags, doesn't seems good to use boolean flags as strings?

ojemuyiwa:
> yup it works fine. at the moment the trainees list either displays all doctors or under notice doctors, there is no other variant. therefore the `underNotice` query parameter as true|false is always sent when the api is called in order to show all|under notice doctors.
> 
> yea i think there could be more filters coming in the future hence i renamed `underNotice` state slice to `filter` to going forward tweaks can be easily made

I agree: my initial question was, this is a parameter i.e. queryParam: { pageNumber:..., underNotice: ...} adding complexity a new object called filter: { underNotice:... } seems boilerplate and less of an improvement? Am I communicating clearly?

omarCreativeDev:
> { underNotice:... } seems boilerplate and less of an improvement? Am I communicating clearly?

@ojemuyiwa im not adding/appending a new object. I have simply renamed the existing `underNotice` property to `filter`

src/app/core/trainee/trainee.service.ts
omarCreativeDev: i agree it would be cleaner @babul3012 but unfortunately the `.set()` method on angular's `HttpParams` class requires the value to be a string

omarCreativeDev:
@babul3012 have updated commit message

src/app/trainees/state/trainees.state.spec.ts
```diff
@@ -107,13 +107,13 @@ describe("Trainees state", () => {
 
   it("should dispatch 'UnderNoticeFilter' and update store", () => {
     store.dispatch(new UnderNoticeFilter());
-    const underNotice = store.selectSnapshot(TraineesState.underNotice);
-    expect(underNotice).toBeTruthy();
+    const filter = store.selectSnapshot(TraineesState.filter);
+    expect(filter).toEqual("underNotice");
```
babul3012, line 111: Can we serve all this strings("underNotice", "allDoctors") from a const, so that it will be cleaner and reusable.
Apologies for not raising this in earlier comments.

ojemuyiwa:
> > { underNotice:... } seems boilerplate and less of an improvement? Am I communicating clearly?
> 
> @ojemuyiwa im not adding/appending a new object. I have simply renamed the existing `underNotice` property to `filter`

Yup and thats what we're discussing... but...

Screenshot 2020-04-21 at 10 24 21
Screenshot 2020-04-21 at 10 23 44

src/app/trainees/state/trainees.state.spec.ts
omarCreativeDev: not a problem @babul3012 I appreciate the feedback, I have converted the string type to an enum instead. pr updated

omarCreativeDev:
> > > { underNotice:... } seems boilerplate and less of an improvement? Am I communicating clearly?
> > 
> > 
> > @ojemuyiwa im not adding/appending a new object. I have simply renamed the existing `underNotice` property to `filter`
> 
> Yup and thats what we're discussing... but...
> 
> Screenshot 2020-04-21 at 10 24 21
> Screenshot 2020-04-21 at 10 23 44

@ojemuyiwa I alphabetised the model so github seems to show lots of changes but i can assure you no new properties were created or removed

PR #58, opened 2020-04-24
feat: trainees date format

omarCreativeDev:
TISNEW-4212

transform dates to correct dd/mm/yyyy format


omarCreativeDev:
fair point, i was initially also thinking about using the pipe in the template but it would have meant lots of ngIf conditional statements would have to be added in the html so went with piping the data once retrieved from the endpoint

omarCreativeDev:
@ojemuyiwa @john12321 pr updated : )

PR #62, opened 2020-04-28
chore: add unit tests to github workflow, rename & optimize script

omarCreativeDev:
- add unit tests to github workflow
- rename & optimize build script

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -25,8 +25,11 @@ jobs:
       - name: Pull down dependencies
         run: npm install
 
+      - name: Run unit tests
+        run: npm run test:ci
+
       - name: Build
-        run: npm run build:shared && npm run build:browsers && ng build --prod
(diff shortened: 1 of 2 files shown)


paulhoang-hee:
I can't really say much on the change (an FE person would be best to ask) but thanks for making an improvement to the pipeline!

.github/workflows/deploy.yml
```diff
@@ -25,8 +25,11 @@ jobs:
       - name: Pull down dependencies
         run: npm install
 
+      - name: Run unit tests
+        run: npm run test:ci
```
ojemuyiwa, line 29: is this "ci" a headless command? or e2e

omarCreativeDev: yea the `test:ci` script essentially runs the unit tests in chrome headless mode and exits once complete and doesn't watch files for changes

in package.json it runs `ng test --watch=false --browsers=ChromeHeadless --code-coverage=true`

PR #67, opened 2020-05-07
Tisnew 4404 cognito login

ojemuyiwa:


src/app/core/auth/auth.service.ts
```diff
@@ -0,0 +1,26 @@
...
+  providedIn: "root"
+})
+export class AuthService {
+  constructor() {}
+
+  currentSession(): Observable<CognitoUserSession> {
+    return from(Auth.currentSession());
+  }
+
+  signIn(): Promise<any> {
```
babul3012, line 17: Why do we need signIn and signOut methods? Hosted UI would take care of that, isn't it?
And even if we need signIn, why federatedSignIn? Isn't that used for social logins like Google / Fb etc.

babul3012: I can see signOut is used and makes sense atm.

babul3012:
In general, a fresh PR with new feature shouldn't have 'fix' commits until if you fix an existing bug in the code. Code smells and duplications should be squashed with original commits.

src/app/core/auth/auth.service.ts
ojemuyiwa: SignIn and SignOut are aws-amplify methods I'm calling within our service which: build external hosted url, redirect and store needed keys in localstorage. SignOut clears storage and cookies then redirects to hosted UI and clears session.

ojemuyiwa: see auth.initializer.ts and status-bar.component.ts for usage

ojemuyiwa:
> In general, a PR with new feature shouldn't have 'fix' commits until if you fix an existing bug in the code. Code smells and duplications should be squashed with original commits.

ok, will do.

Judge40:
> In general, a fresh PR with new feature shouldn't have 'fix' commits until if you fix an existing bug in the code. Code smells and duplications should be squashed with original commits.

I'm actually okay with not doing it in this case, the sonarcloud analysis is essentially a code review and I'm usually happy for reviews to be resolved in a new commit as it makes it easier to review the changes.

It can get complicated if people aren't used to interactive rebase and fixup too, as it doesn't always make sense to add the fixes in to the latest commit.

src/app/core/auth/auth.resolver.ts
```diff
@@ -0,0 +1,26 @@
...
+  ActivatedRouteSnapshot,
+  RouterStateSnapshot
+} from "@angular/router";
+import { Observable } from "rxjs";
+import { Injectable } from "@angular/core";
+import { AuthService } from "./auth.service";
+import { catchError } from "rxjs/operators";
+
+@Injectable()
+export class AuthResolver implements Resolve<any> {
```
omarCreativeDev, line 12: is this resolver needed as the same check is being done by the `initializeApplication()` within `auth.initalizer.ts`?

ojemuyiwa: Short answer: No. Long answer. If authenticated and role !== RevalAdmin we'd need to use new route e.g. error/notauthourised etc. We can delete as its not used / imported anywhere but guaranteed we'd write this again. Hope that makes sense?

omarCreativeDev: ah im with you now. thanks for explaining. as discussed lets take this out for now and tackle once we implement roles logic

ojemuyiwa: deleted as requested

PR #80, opened 2020-05-18
feat: UI optimisation

ojemuyiwa:
removed sticky navigation
revised desktop navigation from 2 rows to one row
fixed bug with skip to maincontent link

TISNEW-4577

### DIFF
--- a/src/app/shared/main-navigation/mat-main-nav/mat-main-nav.component.ts
+++ b/src/app/shared/main-navigation/mat-main-nav/mat-main-nav.component.ts
@@ -1,16 +1,17 @@
-import { Component, Input } from "@angular/core";
+import { Component, Input, OnInit } from "@angular/core";
 import { BreakpointObserver, Breakpoints } from "@angular/cdk/layout";
 import { MatIconRegistry } from "@angular/material/icon";
 import { Observable } from "rxjs";
-import { map, shareReplay } from "rxjs/operators";
+import { map, shareReplay, filter } from "rxjs/operators";
 import { DomSanitizer } from "@angular/platform-browser";
+import { Router, NavigationEnd } from "@angular/router";
 
 @Component({
   selector: "app-mat-main-nav",
   templateUrl: "./mat-main-nav.component.html",
   styleUrls: ["./mat-main-nav.component.scss"]
 })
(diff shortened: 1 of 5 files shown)


src/app/trainees/trainees.component.html
```diff
@@ -1,8 +1,8 @@
 <app-trainee-filters></app-trainee-filters>
 
-<div class="d-grid trainees-top-section align-items-center">
+<div class="d-grid trainees-top-section align-items-center sticky-toolbar">
```
ojemuyiwa, line 3: maybe desktop only? takes up real-estate on mobile.
Screenshot 2020-05-18 at 13 43 43

src/app/shared/main-navigation/mat-main-nav/mat-main-nav.component.ts
```diff
@@ -1,31 +1,45 @@
...
-    private domSanitizer: DomSanitizer
+    private domSanitizer: DomSanitizer,
+    private router: Router
   ) {
     this.addNHSIcon();
   }
+  ngOnInit(): void {
+    this.router.events
+      .pipe(filter((e) => e instanceof NavigationEnd))
+      .subscribe(() => {
```
omarCreativeDev, line 35: i think you'll have to handle the unsubscription otherwise the observable subscription will be left open..

src/app/trainees/trainees.component.html
omarCreativeDev: i think its a good idea for the sticky content to be present on both mobile and desktop. the padding can be tweaked for mobile view to make it more tighter but this looks good

src/app/shared/main-navigation/mat-main-nav/mat-main-nav.component.ts
ojemuyiwa: done

src/app/trainees/trainees.component.html
ojemuyiwa: yeah we'll look at padding across the board. maincontent padding for mobile, pad, desktop to have same as toolbar and then max-width for the whole UI container at some point.

PR #82, opened 2020-05-19
refactor: create shared state

omarCreativeDev:
TISNEW-4523

refactor to create a shared state so it can be extended and consumed by both trainees list and concerns list

### DIFF
--- a/src/app/concerns/concerns.interfaces.ts
+++ b/src/app/concerns/concerns.interfaces.ts
@@ -1,6 +1,20 @@
 import { IGetRecordsResponse } from "../shared/records/records.interfaces";
 
-// TO DO: to be defined once BE is ready
+// TO DO: to be defined
 export interface IGetConcernsResponse extends IGetRecordsResponse {
   concernsInfo: any;
+}
+
+// TO DO: to be defined
+export interface IConcern {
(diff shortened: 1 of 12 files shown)


src/app/concerns/concerns.interfaces.ts
```diff
@@ -1,6 +1,21 @@
...
+  doctorFirstName: string;
+  doctorLastName: string;
+  doctorStatus: string;
+  gmcReferenceNumber: string;
+  lastUpdatedDate: string;
+}
+
+// TO DO: to be defined
+export enum ConcernsFilterType {
+  UNDER_NOTICE = "underNotice",
```
ojemuyiwa, line 19: Just a *note: concerns are opened or closed. Concerns are raised anonymously if a doctor may be suspected of malpractice or abuse. database and service is different to. recommendation/revalidation. As previously suggested if initially separated the refactor can come about from combining common code as there are still unknowns. Speak with Ify she has it all mapped out in her head and could give insight as to filters, functionality etc. https://marvelapp.com/4dcc048/screen/68143896

omarCreativeDev: cheers good spot, completely forgot to do this. pr updated : )

PR #112, opened 2020-05-26
Fix nginx config2

ojemuyiwa:
re-added headers requested by dev-ops and put location tags inside server directive

### DIFF
--- a/nginx/nginx.conf
+++ b/nginx/nginx.conf
@@ -10,18 +10,52 @@ http {
         ~image/                    max;
     }
 
-server {
-    listen 80;
-    server_name _;
-    expires $expires;
-    index index.html;
-    root /usr/share/nginx/html;
-    include /etc/nginx/mime.types;
+    server {
+        listen 80;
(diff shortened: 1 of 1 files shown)


nginx/nginx.conf
```diff
@@ -10,18 +10,52 @@ http {
...
             try_files $uri $uri/ /index.html;
             add_header Cache-Control "no-store, no-cache, must-revalidate";
         }
-    }
 
+        # Media: images, icons, video, audio, HTC
+        location ~* \.(?:jpg|jpeg|gif|png|ico|cur|gz|svg|svgz|mp4|ogg|ogv|webm|htc)$ {
+            # expires 1M;
+            add_header Cache-Control "public";
+        }
```
johnsimmons1980, line 53: is this going to cause it to crash again?

ojemuyiwa: So these location tags are now inside server { } tag. they were outside before which was the error Liban posted the error.

PR #126, opened 2020-05-27
Fix broken unit tests

ojemuyiwa:


omarCreativeDev:
this is excellent work @ojemuyiwa as this intermittent broken test/s issue was hard to debug so thanks for sorting this

PR #139, opened 2020-05-28
feat: workflow for auto building, linting and running tests

ojemuyiwa:
mainly for dependabot PR's


.github/workflows/deploy.yml
```diff
@@ -1,10 +1,14 @@
...
 on:
+  # Trigger the workflow on push or pull request,
+  # but only for the master branch
   push:
     branches:
       - master
-
+  pull_request:
+    branches:
+      - "**"
```
Judge40, line 11: Can be simplified to `on: push`

.github/workflows/deploy.yml
```diff
@@ -34,32 +38,11 @@ jobs:
...
-        env:
-          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
-          ECR_REPOSITORY: tis-revalidation-v2
-          IMAGE_TAG: ${{ github.sha }}
-        run: |
-          docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG .
-          docker push $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG
-          docker tag $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG $ECR_REGISTRY/$ECR_REPOSITORY:latest
-          docker push $ECR_REGISTRY/$ECR_REPOSITORY:latest
-
```
Judge40, line 58: These steps still need to be done when deploying.

.github/workflows/deploy.yml
ojemuyiwa: Yup, just know for sure quite a lot of pushes will fail the linting and tests (at least on my part)

.github/workflows/deploy.yml
ojemuyiwa: these are the steps done on push to master only

.github/workflows/deploy.yml
Judge40: It allows feature branch results to be pushed up to sonarcloud (when that is fully added in) which is a good thing in my experience. It's useful to see what going on in my branch before I raise the PR.

ojemuyiwa: so basically on pull_request (mainly for dependabot) or after we're sure build linting and tests are working on local branches, we'll do the auto build. Then on push to master (merge from PR) we do the complete + deploy steps. Else on push alone we'd be running multiple failed build processes on Git resources. Hope that makes sense?

.github/workflows/deploy.yml
ojemuyiwa: line 45 ` if: github.ref == 'refs/heads/master' ` you gave me the syntax to run on master branch

.github/workflows/deploy.yml
Judge40: One could make the arguement that all commits should build and pass tests anyway.
Using `on: push` brings it in line with all other repository behaviour, including Jenkins based pipelines, so is the right trigger to use.

.github/workflows/deploy.yml
Judge40: But you haven't added those steps in to the the `deploy` job, so even on master the docker image won't be built anymore.

ojemuyiwa: ohhhh snaap oops

.github/workflows/deploy.yml
ojemuyiwa: true, but being an argument, i'm definitely one that doesn't push code up on complete, but rather as I go along e.g. html templates first before logic is coded. So running tests on push will fail 99% of the time. I respectfully agree to disagree. being that it's done in every other repo doesn't quite make it right for this repo

.github/workflows/deploy.yml
ojemuyiwa: steps added

.github/workflows/deploy.yml
```diff
@@ -56,25 +66,19 @@ jobs:
...
-  deploy:
-    name: Deploy
-    needs: build
-    runs-on: ubuntu-latest
-
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v2
+      # - name: Checkout
+      #   uses: actions/checkout@v2
```
Judge40, line 70: You still need to do checkout, it should also be before the docker build step or the Dockerfile won't be accessible.

.github/workflows/deploy.yml
```diff
@@ -56,25 +66,19 @@ jobs:
...
+      #     aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      #     aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      #     aws-region: eu-west-2
 
-      - name: Log in to Amazon ECR
-        id: login-ecr
-        uses: aws-actions/amazon-ecr-login@v1
+      # - name: Log in to Amazon ECR
+      #   id: login-ecr
+      #   uses: aws-actions/amazon-ecr-login@v1
```
Judge40, line 81: Remove this as it duplicates the above AWS credential/login steps, no need to leave commented code.

.github/workflows/deploy.yml
Judge40: I'll clarify as I'm not sure if it ever got written up in the handbook or not. Having commits pass tests was something that was agreed within the TIS team previously, there are obviously unavoidable exceptions but they should be minimized.

Pushing commits with known failures can complicate debugging of issues, for example if I wanted to use `git bisect` and automatically use test success/failure as a guide then the known failing commit will stop the process and require manual intervention to determine whether it is a false positive or not. If every commit fails then the tool is much more time consuming to use, if not rendered useless.
Squashing commits when merging the PR isn't a solution to that problem either as we prefer unsquashed merge commits (to master) to maintain history and allow easy reversion of feature branches.

You might not feel that it is right for this repo but I can't see any reason to diverge other than affecting your personal workflow, which goes against the team's standard practice.

I'm happy for you to widen the reviewers and get more feedback from the wider team if you wish, but at this point I won't approve it as is.

ojemuyiwa: naa, its fine, build and run tests on '**' all push

.github/workflows/deploy.yml
ojemuyiwa: done

.github/workflows/deploy.yml
ojemuyiwa: done

PR #142, opened 2020-05-29
refactor: simplify records service

omarCreativeDev:
TISNEW-4620

simplify records service methods by passing in relevant action to baseclass so that future new list pages can be easily created.

- also relocated trainees list column data so that it can be reused
- fixed unit tests

### DIFF
--- a/src/app/shared/records/record-list/record-list.component.spec.ts
+++ b/src/app/shared/records/record-list/record-list.component.spec.ts
@@ -55,12 +55,14 @@ describe("RecordListComponent", () => {
   beforeEach(() => {
     fixture = TestBed.createComponent(RecordListComponent);
     component = fixture.componentInstance;
-    recordsService.getAction = Get;
-    recordsService.sortAction = Sort;
-    recordsService.resetSortAction = ResetSort;
-    recordsService.paginateAction = Paginate;
-    recordsService.resetPaginatorAction = ResetPaginator;
-    recordsService.searchAction = Search;
+    recordsService.setActions(
+      Get,
+      Sort,
+      ResetSort,
+      Paginate,
+      ResetPaginator,
+      Search
+    );
     fixture.detectChanges();
   });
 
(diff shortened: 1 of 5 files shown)


src/app/shared/records/record-list/record-list.component.spec.ts
```diff
@@ -47,6 +55,12 @@ describe("RecordListComponent", () => {
   beforeEach(() => {
     fixture = TestBed.createComponent(RecordListComponent);
     component = fixture.componentInstance;
+    recordsService.getAction = Get;
```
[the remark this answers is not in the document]
omarCreativeDev: cheers for spotting, have updated pr

src/app/shared/records/services/records.service.ts
```diff
@@ -84,50 +84,50 @@ export class RecordsService {
...
   }
 
-  public sort(active: string, direction): void {
-    if (this.isTraineesState) {
-      this.store.dispatch(new SortTrainees(active, direction));
-    } else if (this.isConcernsState) {
-      this.store.dispatch(new SortConcerns(active, direction));
+  public sort(active: string, direction): Observable<any> {
+    if (!this.sortAction) {
+      throw new Error("sortActionFunction must be defined");
```
babul3012, line 96: Could you please explain this? Why do we want to throw error?
What will happen in the page?

src/app/trainees/trainee-list/trainee-list.component.ts
```diff
@@ -20,58 +29,7 @@ export class TraineeListComponent extends RecordListComponent
...
-      name: "admin",
-      enableSort: false
-    },
-    {
-      label: "Last updated",
-      name: "lastUpdatedDate",
-      enableSort: true
-    }
-  ];
+  public columnData = COLUMN_DATA;
```
babul3012, line 32: Where are we using this variable? I didn't find any references in this file.

PR #148, opened 2020-06-02
Ui history form and route

ojemuyiwa:
UI enhancements and agreement legal notice prepared for data-binding


src/app/app.module.ts
```diff
@@ -51,7 +52,7 @@ Amplify.configure(AWS_CONFIG);
       provide: APP_INITIALIZER,
       useFactory: initializeApplication,
       multi: true,
-      deps: [AuthService]
+      deps: [AuthService, Router]
```
omarCreativeDev, line 55: just for my own understanding is the Router added here so that amplify can do the redirect?

PR #156, opened 2020-06-03
refactor: cleanup and remove redundant scss

omarCreativeDev:
cleanup and remove redundant scss. this was outstanding from when we switched over to using material over nhsuk toolkit


src/scss/nhsuk/_index.scss
```diff
@@ -0,0 +1,5 @@
+// Settings
+@import "src/scss/nhsuk/colours";
```
ojemuyiwa, line 2: do we need the full-path? i.e would import 'colours' work? ignore if it all imports as is anyways

omarCreativeDev: good spot cheers! have resolved

PR #176, opened 2020-06-08
Plugin save and submit endpoints to fe buttons

ojemuyiwa:


package.json
```diff
@@ -5,7 +5,7 @@
     "ng": "ng",
     "start": "npm run build:misc && ng serve",
     "start:local": "npm run build:misc && ng serve --configuration=local",
-    "build": "ng build",
+    "build": "node --max_old_space_size=4096 node_modules/@angular/cli/bin/ng build",
```
ojemuyiwa, line 8: Hey Omar let me know what you think about this angular suggested hack, makes the build faster i've noticed

omarCreativeDev: can i ask what problem this is trying to address? a build output size issue? is an issue with angular itself?

omarCreativeDev: also would this also be required in the `github > deploy.yml` file?

ojemuyiwa:
> pr looks fine, perhaps its possible to break down some of the components further in to child components? `RevalidationHistoryComponent` might be a good candidate to split (if possible) as there's a lot happening in there

Absolutely!!!! You're 101% correct! Just has its few challenges and would've added at least 3 days to the item. So like your refactor i've left it for the slow days. Only did the table, but will remove each form so its split up nicely

PR #184, opened 2020-06-12
chore: outdated dependencies

ojemuyiwa:


omarCreativeDev:
Hi @ojemuyiwa there is a conflict with `supported.browsers.js` - i believe this file is now untracked + part of git ignore on master so perhaps you just have to rebase from master to fix

PR #185, opened 2020-06-15
chore: cleanup environment files

omarCreativeDev:
TISNEW-4756

- cleanup environment files by alphabetising
- renamed `environment.local` file to mocky and ensured mocky hosted endpoints are in this file - and nowhere else
- cleaned up npm tasks to leverage parameters via `--`
- updated `REAME.md` to reflect


package.json
```diff
@@ -4,13 +4,13 @@
   "scripts": {
     "ng": "ng",
     "start": "npm run build:misc && ng serve",
-    "start:local": "npm run build:misc && ng serve --configuration=local",
+    "start:mocky": "npm run start -- --configuration=mocky",
```
babul3012, line 7: why are we using the mocky configuration?

omarCreativeDev: purely so that during local development FE dont have to reply on BE updates to an environment and so that FE can quickly stub out a BE endpoint. or perhaps if an environment is down etc is unblocks the FE

PR #188, opened 2020-06-15
refactor: setup sub routes

omarCreativeDev:
TISNEW-4756

first step of breaking down `RecommendationHistoryComponent` is to create some sub routes so that the amount of logic and code during each user journey step is kept to a minimum. will also aid in reusing/extending the code on other similar details pages


src/app/recommendation/recommendation-routing.module.ts
```diff
@@ -15,6 +17,18 @@ const routes: Routes = [
         component: RecommendationHistoryComponent,
         data: { title: "Recommendation history" },
         resolve: { recommendation: RecommendationResolver }
+      },
+      // TODO double check if a guard and/resolver is needed on this route
```
ojemuyiwa, line 21: Looking at the code i think the routing strategy may work better if the create/edit and confim components were children of the history route.

ojemuyiwa: they would both use the logic for displaying create edit recommendation button as guard, and the form would have to be valid on create route for submit route to allow activation.

omarCreativeDev: each route is essentially performing a unique job in the user flow i.e list, create and submit so im not sure what the benefits of nesting the create & submit routes under history would be? especially as the buttons are binded to a slice on state management...

absolutely agreed that the create/edit and submit routes need to be protected with guard/s and/or resolver/s. I added some TODO's in the routing module just so that we dont forget to address

ojemuyiwa: You already answered your question. Each route performing a unique job in user flow. Data loads once at the moment. Parent history route resolver gets this data in order to display the history. edit/create uses the historical data, submit: submits the historical data. Hence Recommendation History is parent and will fetch the data once when needed. Even if you hit child route e.g.  `recommendation/{gmcId}/create` the parent route resolves the data before child logic.

ojemuyiwa: Screenshot 2020-06-15 at 17 35 07
furthermore:
with router-outlet in history component. All child routes will only load in circled space.

ojemuyiwa: so `children:[{"", historicaltable}, {"create", createComponent}, {"edit", createComponent}, "submit", submitComponent}]` one resolver on parent and guards only on children.

omarCreativeDev: ah yes i see your point now with the `RecommendationResolver` being required to run on all 3 routes. wouldn't it be better to move the resolver to the parent i.e `RecommendationComponent` instead of creating a grandchild relationship if that makes sense? pr updated

ojemuyiwa: Yup, this works perfectly as well, many ways to skin a cat. you may even want a details holder component that's shared and uses same instance for recommendation details, connections details, and concerns details. this component will have `<app-nav-bar>` on the top and `<app-trainee-details>` on the left. and would load once via imports/exports in shared/details-module. so all sub modules share a large chunk of code.

omarCreativeDev: this is a very good idea, will take a look shortly

PR #208, opened 2020-06-25
chore: update browser-list

ojemuyiwa:


omarCreativeDev:
@ojemuyiwa there is a small conflict in the package-lock.json file but this is good to go in otherwise

PR #214, opened 2020-06-30
Concerns details view

ojemuyiwa:


src/app/concern/concern.component.html
```diff
@@ -1 +1,171 @@
-<p>concern works!</p>
+<div class="table-container" *ngIf="concernHistory$ | async as concernHistory">
```
omarCreativeDev, line 1: this seems really similar to RecommendationTableComponent. is it possible to create a generic historytable component so it can be reused on both places and just pass in the datasource, column names etc?

src/app/concern/concern.component.ts
```diff
@@ -1,7 +1,60 @@
...
 @Component({
   selector: "app-concern",
-  templateUrl: "./concern.component.html"
+  templateUrl: "./concern.component.html",
+  styleUrls: [
+    "./concern.component.scss",
+    "../../app/shared/details/details.table.scss"
+  ],
+  animations: [
+    trigger("detailExpand", [
```
omarCreativeDev, line 22: can this animation be extracted in to a common place and re used?

PR #215, opened 2020-06-30
get admin users list

omarCreativeDev:
TISNEW-4829

feat: get admin users list

- fetch admins list via aws cognito sdk api and store on `admins` state slice
- invoke `Get` admins action from app component as the admins list is to be used across all summary & details pages
- extend default angular webpack config so that environment variables can be consumed for AWS `secretAccessKey` & `AWS_ACCESS_KEY_ID`
- updated `README.md`

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -10,6 +10,7 @@ import { AnalyticsModule, HotJarModule } from "hee-shared";
 import { AppRoutingModule } from "./app-routing.module";
 import { AppComponent } from "./app.component";
 import { AWS_CONFIG } from "./core/auth/aws-config";
+import { AdminsState } from "./shared/admins/state/admins.state";
 import { MainNavigationModule } from "./shared/main-navigation/main-navigation.module";
 import { MaterialModule } from "./shared/material/material.module";
 import { SharedModule } from "./shared/shared.module";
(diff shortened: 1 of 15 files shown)


src/app/shared/admins/services/admins.service.ts
```diff
@@ -0,0 +1,59 @@
...
+@Injectable({
+  providedIn: "root"
+})
+export class AdminsService {
+  constructor(private store: Store, private snackBar: MatSnackBar) {}
+
+  public get listUsersInGroupRequest(): ListUsersInGroupRequest {
+    return {
+      UserPoolId: environment.awsConfig.userPoolId,
+      GroupName: "site-admin-group"
```
ojemuyiwa, line 25: this looks like it could be re-usable function which accepts `groupName` as parameter. Or could be hard-coded in `getAdminUsers` as the function name implies

ojemuyiwa: *Note group names have been slightly altered and may incorporate multiple rules `https://hee-tis.atlassian.net/wiki/spaces/NTCS/pages/1671495738/Revalidation+-+Access+Permissions`

src/app/app.module.ts
```diff
@@ -42,7 +43,9 @@ Amplify.configure(AWS_CONFIG);
       hotJarSv: environment.hotJarSv,
       enabled: environment.production
     }),
-    NgxsModule.forRoot([], { developmentMode: !environment.production }),
+    NgxsModule.forRoot([AdminsState], {
```
ojemuyiwa, line 46: Is it better forRoot here or in it's own module? looking more like an auth-module may be good to encapsulate all auth-related services, states, functions? just a thought though.

PR #223, opened 2020-07-03
feat: handle row interaction

omarCreativeDev:
TISNEW-4834

handle row interaction across recommendations, concerns & connections screens via checkboxes and disabled navigation to details screen if allocateAdmin mode is activated

- created new `ToggleCheckbox` & `ToggleAllCheckboxes` actions for all 3 screens
- updated state to store `allChecked`, `someChecked` (for materials indeterminate ui state) and `checked` on each item
- created new `allChecked` & `someChecked` selectors
- updates to interfaces and fixed tests


src/app/concerns/state/concerns.actions.ts
```diff
@@ -59,3 +60,11 @@ export class ResetPaginator {
 export class EnableAllocateAdmin extends EnableAllocateAdminPayload {
   static readonly type = `${label} Enable Allocate Admin`;
 }
+
```
babul3012, line 63: These following methods seems to be duplicated in connections as well. Isn't it better to extract them to a common method?

omarCreativeDev: cheers for flagging, yea just looking at this now. am consolidating the common actions in to a common place

PR #255, opened 2020-07-16
feat: consume upload api

omarCreativeDev:
TISNEW-4914

feat: consume upload api

- also extracted common code for opening up a snackbar (material's version of a toaster notification)
- updated unit tests

### DIFF
--- a/src/app/concern/file-uploader/file-uploader.component.ts
+++ b/src/app/concern/file-uploader/file-uploader.component.ts
@@ -1,6 +1,11 @@
 import { Component, ElementRef, OnInit, ViewChild } from "@angular/core";
 import { FormBuilder, FormGroup } from "@angular/forms";
+import { Store } from "@ngxs/store";
+import { Observable } from "rxjs";
+import { SnackBarService } from "../../shared/services/snack-bar/snack-bar.service";
 import { UploadService } from "../services/upload/upload.service";
+import { Upload } from "../state/concern.actions";
+import { ConcernState } from "../state/concern.state";
 
 @Component({
   selector: "app-file-uploader",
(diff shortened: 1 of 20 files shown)


src/app/concern/file-uploader/file-uploader.component.ts
```diff
@@ -38,31 +46,46 @@ export class FileUploaderComponent implements OnInit {
...
-    // const dt = e.dataTransfer;
-    // this.uploadedFiles = dt.files;
-    this.upload();
+  public handleDrop($event: DragEvent): void {
+    const itemList: DataTransferItem[] = Array.from($event.dataTransfer.items);
+    const filesAllowed: string[] = this.acceptedFileTypes.split(",");
+    const droppedFiles: File[] = [];
+    const invalidFiles: string[] = [];
+
+    if (itemList.length) {
```
babul3012, line 55: Wouldn't this throw an error if itemList is null or undefined?

src/app/concern/file-uploader/file-uploader.component.ts
```diff
@@ -38,31 +46,46 @@ export class FileUploaderComponent implements OnInit {
...
+
+      itemList.forEach((i: DataTransferItem) => {
+        if (filesAllowed.includes(i.type)) {
+          droppedFiles.push(i.getAsFile());
+        } else {
+          invalidFiles.push(i.getAsFile().name);
+        }
+      });
+
+      if (invalidFiles.length) {
```
babul3012, line 70: We only have one file at a time and would push the same to droppedFiles / InvalidFiles. Why do need to push them in list and check length? Can't we return from above if.

src/app/concern/file-uploader/file-uploader.component.ts
omarCreativeDev: good spot, have updated pr

src/app/concern/file-uploader/file-uploader.component.ts
```diff
@@ -38,31 +46,46 @@ export class FileUploaderComponent implements OnInit {
     this.dropArea.nativeElement.classList.remove("highlight");
   }
 
-  public handleDrop(e: any): void {
-    // const dt = e.dataTransfer;
-    // this.uploadedFiles = dt.files;
-    this.upload();
+  public handleDrop($event: DragEvent): void {
+    const itemList: DataTransferItem[] = Array.from($event.dataTransfer.items);
+    const filesAllowed: string[] = this.acceptedFileTypes.split(",");
```
babul3012, line 51: Can't we declare the allowedFileTypes as array instead of splitting into an array?

src/app/concern/file-uploader/file-uploader.component.ts
omarCreativeDev: very true, BE will be enabling multiple file upload feature shortly so built the FE with a list in mind : )

babul3012: I would prefer to add the code later as future is very uncertain ;)

src/app/concern/file-uploader/file-uploader.component.ts
omarCreativeDev: good point, makes sense. i will do this now

omarCreativeDev: pr updated @babul3012

PR #317, opened 2020-08-05
Tisnew 5043 create concern

ojemuyiwa:


src/app/concern/concern.interfaces.ts
```diff
@@ -10,6 +10,11 @@ export interface IListFile {
   key: string;
 }
 
+export interface IFileUploadeProgress {
```
omarCreativeDev, line 13: is it possible to fix the spelling of the interface name?

src/app/concern/file-uploader/file-uploader.component.ts
```diff
@@ -24,27 +25,36 @@ export class FileUploaderComponent implements OnInit {
     "text/plain",
     "text/csv"
   ];
+  public acceptedFileSize = 1048576;
```
omarCreativeDev, line 28: i think you should speak to ify about a max file size limit as the requirement in the story is to not have a file size restriction...

src/app/concern/concern.interfaces.ts
ojemuyiwa: Good catch: fixed

src/app/concern/file-uploader/file-uploader.component.ts
ojemuyiwa: Absolutely after chatting on the group: I think Ify is fine with the solution, as she re-worded the text during conversations to 10MB

omarCreativeDev: sounds good

PR #325, opened 2020-08-07
fix: e2e tests

omarCreativeDev:
NO-TICKET

fix: e2e tests

The pr addresses the issue of login & as well as fixes the broken tests.

- The externally hosted aws cognito login screen is filled in and submitted
- the baseUrl has been set to staging as cypress doesn't allow tests to run on different domains. The only other option I could think of was to progmatically login via `cy.request()` but this would expose AWS secrets which is not a good idea

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,3 +1,4 @@
 {
-  "baseUrl": "http://localhost:4200"
+  "baseUrl": "https://stage-revalidation.tis.nhs.uk",
+  "chromeWebSecurity": false
 }
--- a/cypress.json
+++ b/cypress.json
@@ -1,4 +1,3 @@
 {
-  "baseUrl": "https://stage-revalidation.tis.nhs.uk",
   "chromeWebSecurity": false
 }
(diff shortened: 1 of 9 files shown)


cypress.json
```diff
@@ -1,3 +1,4 @@
 {
-  "baseUrl": "http://localhost:4200"
+  "baseUrl": "https://stage-revalidation.tis.nhs.uk",
```
ojemuyiwa, line 2: unsure how this will run on pipeline when hasn't been deployed?

ojemuyiwa: was looking at `https://www.cypress.io/blog/2019/11/20/drastically-simplify-your-testing-with-cypress-github-action/` a while ago? unsure if we may need this

omarCreativeDev: i see your point, i suppose this would mean the feature would have to be deployed first and the e2e tests come after. but not sure if this restriction is a good idea but not sure how to get around the cypress multiple domain restriction...

omarCreativeDev: thats a great article! I've added it to the pipeline ticket TISNEW-4706 (currently in the backlog at the moment). be worthwhile discussing the reval e2e pipeline with the other FEDS

ojemuyiwa: on git unbutu server npm start will launch it locally to the docker-image then cypress will run against it and pass or fail. All prior to deploying to staging

PR #337, opened 2020-08-11
Create concern ui enhancements

ojemuyiwa:
Enhanced UI
Fixed unit tests
Added unit tests


src/app/shared/services/snack-bar/snack-bar.service.ts
```diff
@@ -11,7 +11,7 @@ import {
 export class SnackBarService {
   public horizontalPosition: MatSnackBarHorizontalPosition = "center";
   public verticalPosition: MatSnackBarVerticalPosition = "top";
-  public snackBarDuration = 10000;
+  public snackBarDuration = 100000;
```
omarCreativeDev, line 14: should this be 10000 as opposed to 100000?

ojemuyiwa: Oooops was using to style in console. good spot

PR #339, opened 2020-08-11
fix: admin items subscription

omarCreativeDev:
NO-TICKET

fix: admin items subscription by adding a rxjs filter


src/app/admins/allocate-admin-autocomplete/allocate-admin-autocomplete.component.ts
```diff
@@ -31,7 +37,7 @@ export class AllocateAdminAutocompleteComponent implements OnInit {
 
   public getItems(): void {
     this.items$
-      .pipe(take(1))
+      .pipe(filter(Boolean), take(1))
```
ojemuyiwa, line 40: cool beans, just for my reference what does this do? commented out this line and noticed it works without the take(1)

omarCreativeDev: so take will basically complete a subscription based upon the number you passed in to it, i.e in this case it will take the first emitted value. the bug was the admins list first emission value was null so adding a filter(Boolean) just ignores the null emission so when the next emission happens (an actual admins list) its allowed through and then completes / unsubscribes

PR #351, opened 2020-08-17
refactor: create generic enum key grabber

omarCreativeDev:
TISNEW-5042

refactor: create generic enum key grabber in order to send BE enum keys when invoking api's

- also set responseType to text for add concern api


omarCreativeDev:
@ojemuyiwa as discussed i have updated the pr and have provided the service to the shared module : )

PR #406, opened 2020-09-04
fix: render admin full-name and email mailTo links

ojemuyiwa:
TISNEW-5375

### DIFF
--- a/src/app/admins/admins.pipe.ts
+++ b/src/app/admins/admins.pipe.ts
@@ -11,13 +11,17 @@ export class AdminsPipe implements PipeTransform {
 
   constructor(private store: Store) {}
 
-  transform(email: string): string {
+  transform(email: string, format?: string): string {
     if (email && this.admins) {
       const admin: IAdmin = this.admins.find(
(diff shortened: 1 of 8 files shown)


src/app/admins/admins.pipe.ts
```diff
@@ -11,13 +11,17 @@ export class AdminsPipe implements PipeTransform {
...
 
-  transform(email: string): string {
+  transform(email: string, format?: string): string {
     if (email && this.admins) {
       const admin: IAdmin = this.admins.find(
         (a: IAdmin) => a.username === email
       );
 
-      return admin ? admin.fullName : email;
+      if (format === "email") {
```
babul3012, line 20: Can we make it in single line?
```return admin? (format === "email"? admin.email : admin.fullName) : email;```

ojemuyiwa: done

babul3012 (approved):

babul3012:
I think we would be able to read the name from the token now and use it!

PR #407, opened 2020-09-07
refactor: Admin pipe to use the user information from the token

babul3012:
TISNEW-5375

### DIFF
--- a/src/app/admins/admins.pipe.ts
+++ b/src/app/admins/admins.pipe.ts
@@ -2,28 +2,22 @@ import { Pipe, PipeTransform } from "@angular/core";
 import { Store } from "@ngxs/store";
 import { IAdmin } from "./admins.interfaces";
 import { AdminsState } from "./state/admins.state";
+import { AuthService } from "../core/auth/auth.service";
+import { switchMap } from "rxjs/operators";
 
 @Pipe({
   name: "AdminName"
 })
 export class AdminsPipe implements PipeTransform {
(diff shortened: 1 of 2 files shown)


omarCreativeDev (approved):

src/app/admins/admins.pipe.ts
```diff
@@ -2,28 +2,22 @@ import { Pipe, PipeTransform } from "@angular/core";
 import { Store } from "@ngxs/store";
 import { IAdmin } from "./admins.interfaces";
 import { AdminsState } from "./state/admins.state";
+import { AuthService } from "../core/auth/auth.service";
+import { switchMap } from "rxjs/operators";
```
omarCreativeDev, line 6: @babul3012 could you remove the unused imports from this file such as switchMap etc?

src/app/admins/admins.pipe.ts
```diff
@@ -2,28 +2,22 @@ import { Pipe, PipeTransform } from "@angular/core";
...
-      const admin: IAdmin = this.admins.find(
-        (a: IAdmin) => a.username === username
-      );
-
-      return admin
-        ? format === "email"
-          ? admin.email
-          : admin.fullName
-        : username;
+    if (username) {
```
ojemuyiwa, line 15: this will break the pipe usage in templates and tests.

babul3012: I think this always return the logged in User name, changing this pipe is not a solution.

PR #423, opened 2020-09-10
fix: username in status-bar 

ojemuyiwa:
display email for key-cloak and full name in cognito

TISNEW-5421

### DIFF
--- a/src/app/shared/main-navigation/mat-main-nav/status-bar/status-bar.component.ts
+++ b/src/app/shared/main-navigation/mat-main-nav/status-bar/status-bar.component.ts
@@ -22,7 +22,9 @@ export class StatusBarComponent implements OnInit {
   }
 
   setUsername(): void {
-    this.username = this.authService.userName;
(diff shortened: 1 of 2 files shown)


src/app/shared/main-navigation/mat-main-nav/status-bar/status-bar.component.ts
```diff
@@ -22,7 +22,9 @@ export class StatusBarComponent implements OnInit {
   }
 
   setUsername(): void {
-    this.username = this.authService.userName;
+    this.username = this.authService.userName
```
babul3012, line 25: I guess we always get the username now(email) and very less scope for fullName to return.

ojemuyiwa: no "custom:preferred_username" when login with cognito user. blank

babul3012 (approved):

PR #490, opened 2020-10-29
feat: Load environment variables dynamically

babul3012:
TISNEW-5658


src/main.ts
```diff
@@ -1,13 +1,51 @@
...
+import { Amplify } from "aws-amplify";
 import { AppModule } from "./app/app.module";
-import { environment } from "./environments/environment";
+import { environment as defaultEnvironment } from "./environments/environment";
+
+(async () => {
+  const envApiResponse = await fetch("/api/environment");
+  const envName = await envApiResponse.json();
+
+  const response = await fetch("./assets/environments/environment.json");
```
Judge40, line 11: Could we concat the `envName` to get values from the standard `environment.<env-name>.ts` files at all? Or is that more effort than its worth?

Judge40 (approved):

PR #543, opened 2020-11-26
fix: changed the admin value to logged in user full name

jayanta2018:
changed the admin value to logged in user full name
It was originally email address

TISNEW-5729


jayanta2018:
this pr is not needed

PR #618, opened 2021-01-29
feat: enable exception search

DorisWongNHS:
enable search function for connection exception
and set the search label as  "Enter GMC no"

TIS21-264

### DIFF
--- a/src/app/connections/state/connections.state.ts
+++ b/src/app/connections/state/connections.state.ts
@@ -148,7 +148,8 @@ export class ConnectionsState extends RecordsState {
       filter: action.filter,
       disableSearchAndSort:
         action.filter !== ConnectionsFilterType.ALL &&
-        action.filter !== ConnectionsFilterType.HIDDEN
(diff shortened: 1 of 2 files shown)


src/app/connections/state/connections.state.ts
```diff
@@ -148,7 +148,8 @@ export class ConnectionsState extends RecordsState {
       filter: action.filter,
       disableSearchAndSort:
         action.filter !== ConnectionsFilterType.ALL &&
-        action.filter !== ConnectionsFilterType.HIDDEN
+        action.filter !== ConnectionsFilterType.HIDDEN &&
+        action.filter !== ConnectionsFilterType.EXCEPTIONS_QUEUE
```
john12321, line 152: Nothing wrong with this but if we add more to this list it's probably easier/cleaner to filter types to disable rather than those '!=='

john12321 (approved):
Nice work, @DorisWongNHS !

PR #776, opened 2021-09-06
fix: recommendation comment handle html tags

DorisWongNHS:
TIS21-2009


CaiWillis:
Just to check, does the comment always come from a trusted source? I know angular's innerHtml does some sanitization bt it could still be an xss risk depending on who can make comments?

jayanta2018 (commented):
Is it "connection" or comment in the title? Just wondering @DorisWongNHS

CaiWillis (approved):

PR #785, opened 2021-09-28
fix:add quotes to correct font concatenation

StevenHee:
The NHS and fallback font strings are concatenated which was resulting in a font name Frutiger W01Arial.  Fonts now wrapped in quotes to correct this.


jayanta2018 (approved):
Many thanks @StevenHee  for taking time to look into Revalidation FE repo

PR #803, opened 2021-11-05
fix: remove make recommendation from reval admin role

jayanta2018:
There are 3 roles in revalidation arena. They are "RevalApprover", "RevalAdmin" and "RevalObserver". Reval approver can read, write and make recommendation, Reval admin can read and write but should not be allowed to make recommendation. This fix will take care of that.

Reval Observer has only read permission, which has already been handled in the app. 

For more info, please refer to this page
https://hee-tis.atlassian.net/wiki/spaces/NTCS/pages/1671495738/Revalidation+-+Access+Permissions

TIS21-2272

### DIFF
--- a/src/app/recommendation/create-recommendation/create-recommendation.component.html
+++ b/src/app/recommendation/create-recommendation/create-recommendation.component.html
@@ -85,7 +85,7 @@
     >
       Save draft
     </button>
-    <button
+    <button *ngIf="isRevalApprover"
       type="submit"
       mat-button
       mat-raised-button
--- a/src/app/core/auth/auth.service.ts
+++ b/src/app/core/auth/auth.service.ts
@@ -4,7 +4,7 @@ import { CognitoUserSession } from "amazon-cognito-identity-js";
 import { Auth } from "aws-amplify";
 import { from, Observable } from "rxjs";
 import { catchError, tap } from "rxjs/operators";
-import { ADMIN_ROLES, LONDON_DBCS } from "src/environments/constants";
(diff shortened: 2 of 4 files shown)


src/app/recommendation/create-recommendation/create-recommendation.component.html
```diff
@@ -85,7 +85,7 @@
     >
       Save draft
     </button>
-    <button
+    <button *ngIf="!isRevalAdmin"
```
CaiWillis, line 88: Just wondering based on the requirements should it instead be ngIf="isRevalApprover" (and replace the bit in the service.ts with a includes("RevalApprover") - just because it's an exclusive feature for that role (Observers shouldn't be able to see this button either right?)

jayanta2018: Thanks @CaiWillis . Nice suggestion. I have mage changes accordingly

DorisWongNHS:
It looks good to me. Just a small discussion, the existing FE code is putting the role into the constants.ts so the permission setting can be shared to use in different places of the FE. [https://github.com/Health-Education-England/TIS-REVALIDATION-V2/blob/master/src/environments/constants.ts](url). Just think do we consider doing in that way?

jayanta2018:
> It looks good to me. Just a small discussion, the existing FE code is putting the role into the constants.ts so the permission setting can be shared to use in different places of the FE. [https://github.com/Health-Education-England/TIS-REVALIDATION-V2/blob/master/src/environments/constants.ts](url). Just think do we consider doing in that way?

Thanks @DorisWongNHS I have made changes. Please have a look.

DorisWongNHS (approved):

PR #820, opened 2021-12-13
chore: hide concern and connection from main and navigation menu

jayanta2018:
TIS21-2021

- hiding from main menu by changing the "type" to 9, so that desktop-menu.component.html file's logic does not match
- hiding from the navigation bar by commenting the code, we will need this code later

### DIFF
--- a/src/app/shared/main-navigation/mat-main-nav/menu-items.const.ts
+++ b/src/app/shared/main-navigation/mat-main-nav/menu-items.const.ts
@@ -93,17 +93,17 @@ export const menuItems: IMenuItem[] = JSON.parse(
                 "description": ""
             },
             {
-                "type": 0,
+                "type": 9,
                 "route": "/concerns",
                 "name": "Concerns",
                 "description": ""
             },
             {
-                "type": 0,
+                "type": 9,
                 "route": "/connections",
                 "name": "Connections",
                 "description": ""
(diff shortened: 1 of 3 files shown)


src/app/shared/main-navigation/mat-main-nav/menu-items.const.ts
```diff
@@ -93,17 +93,17 @@ export const menuItems: IMenuItem[] = JSON.parse(
...
                 "description": ""
             },
             {
-                "type": 0,
+                "type": 9,
                 "route": "/connections",
                 "name": "Connections",
                 "description": ""
-            }                    
+            }
```
Judge40, line 106: Can these be removed too?

jayanta2018: I tried to run the programme by removing these. Unfornately it did not work without these. That's why I was bound to change the value from 0 to 9
Not sure if there is other way.

jayanta2018: Sorry @Judge40 It wa my bad understanding. I was missing something yesterday. Thanks to @CaiWillis It is now resolved.
I have removed the concern and connection.

Judge40 (approved):

PR #841, opened 2022-01-21
fix: use environment specific bucket for concerns

JosephKelly:
I assume that this is a bit of copy pasta rather than intentional.

NO-CARD


jayanta2018 (approved):
Looks aligned with prod pool and s3 bucket. Nice

PR #842, opened 2022-01-21
chore: use preferred_username isntead of custom claim

CaiWillis:


JosephKelly (approved):
All good but **think the user pool / lambda needs to be updated before this is merged**

PR #849, opened 2022-02-02
fix: change "CCT date" label to "Curriculum end date"

DorisWongNHS:
TIS21-2611


jayanta2018 (approved):

src/app/records/record-list/record-list.component.scss
```diff
@@ -40,7 +40,7 @@
       width: 100px
     ),
     cctDate: (
-      width: 110px
+      width: 130px
```
JosephKelly, line 43: Proper @Health-Education-England/frontend / UI Design question but what does this do for the target device width?

src/app/recommendations/constants.ts
```diff
@@ -6,7 +6,7 @@ export const COLUMN_DATA: [string, string, boolean][] = [
   ["TIS Status", "doctorStatus", false],
   ["Programme name", "programmeName", false],
   ["Programme membership type", "programmeMembershipType", false],
-  ["CCT date", "cctDate", false],
+  ["Curriculum end date", "cctDate", false],
```
JosephKelly, line 9: Can we change `cctDate`? Continuing with that name is likely to confuse "future Pepe".

JosephKelly (changes requested):

src/app/recommendations/constants.ts
DorisWongNHS: Thanks @JosephKelly, it's a good shout. Was thinking about it. In order to make it consistent, we may need to update all names in TCS, Recommendation, Concerns, Core, Integration as well. 
Will make PRs for the backend shortly. Then we merge the frontend as the last one.

src/app/records/record-list/record-list.component.scss
DorisWongNHS: This change increases the width of the "Curriculum end date column". As the column title has been changed longer now, the width needed to be longer so it won't break to another line. I believe the whole table width would be increased by 20px, but with my local testing, it is still fitting on the same page without needing to scroll on my screen. For the mobile setting, the table would need to be scroll originally anyway.

jayanta2018 (approved):
LGTM

JosephKelly (approved):

PR #919, opened 2022-06-15
Fix package vulnerabilities and upgrade Angular to v13

StevenHee:
Key tasks:-

- Remove `HEEShared` child project and associated references
- Add GA4 and HotJar script tag
- Upgrade angular sequentially to v13
- Update node packages where possible
- Remove "Solution Style" `tsconfig` and set as default to fix error

Apologies for the number of files changed but apart from some CSS tweaks, there are no actual code changes.

TICKET: https://hee-tis.atlassian.net/browse/TIS21-3081


src/app/app-routing.module.ts
```diff
@@ -55,7 +55,7 @@ const routes: Routes = [
 ];
 
 @NgModule({
-  imports: [RouterModule.forRoot(routes)],
+  imports: [RouterModule.forRoot(routes, { relativeLinkResolution: 'legacy' })],
```
hihilary, line 58: Should we fix the routerLink later to use default config 'corrected'?

StevenHee: Had a look and we don't have any relative router links with double dot notation so can update this to `corrected` now. Passes unit and e2e tests, for what that's worth!

PR #926, opened 2022-06-24
Enable Cypress cross domain tests and link to Cypress dashboard for reports

StevenHee:
The E2E tests are failing in the pipeline which I suspect is due to Cognito hosted authentication handled on a separate domain, although there is no way to view the reports or videos. To resolve I have enabled this via a Cypress experimental flag and  connected to Cypress Dashboard to view the reports in the immediate term since the entry tier is free but also investigate as a possible solution for storing reports going forward.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,8 @@
     "lint": "ng lint",
     "lint:styles": "stylelint \"src/**/*.scss\"",
     "cypress": "cypress run --browser chrome",
-    "cypress-open": "cypress open",
+    "cypress:local": "cypress run --browser chrome --config baseUrl=http://localhost:4200",
+    "cypress:open": "cypress open --config baseUrl=http://localhost:4200",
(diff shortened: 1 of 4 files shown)


hihilary (approved):

package.json
```diff
@@ -13,7 +13,8 @@
     "lint": "ng lint",
     "lint:styles": "stylelint \"src/**/*.scss\"",
     "cypress": "cypress run --browser chrome",
-    "cypress-open": "cypress open",
+    "cypress:local": "cypress run --browser chrome --config baseUrl=http://localhost:4200",
+    "cypress:open": "cypress open --config baseUrl=http://localhost:4200",
```
JosephKelly, line 17: I couldn't see from a quick DDG but am used to tools where the command line takes precedence over ENV_VARs & config files.  Please let me know / update as needed

PR #932, opened 2022-06-28
Relocate supported.browsers file and change browser versions

StevenHee:
The link to the file `supported.browsers.js` was not working on application build. This was due to it being located in the root folder and the filename gets hashed on build, possible change following upgrade of Angular. Relocating the file to the assets folder avoids the hashing.

Upgrade the browser used for E2E in pipeline to comply with `.browserlistrc` settings. Also fix Chrome version to >= 100 to avoid E2E failing on future runs


hihilary (approved):

jayanta2018 (dismissed):

StevenHee:
After some research, I have fixed by updating a setting in `angular.json` which now seems to build without hashing `supported.browsers.js`. I have therefore reverted the location to the root folder.

jayanta2018 (approved):

PR #936, opened 2022-07-04
Feat/add submission due date

StevenHee:
https://hee-tis.atlassian.net/browse/TIS21-2777

### DIFF
--- a/src/scss/utilities/_utilities.scss
+++ b/src/scss/utilities/_utilities.scss
@@ -1,5 +1,11 @@
 @import "~nhsuk-frontend/packages/core/settings/colours";
 // re-usable styles and mixins
+
+h2 {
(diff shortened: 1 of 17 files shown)


src/scss/utilities/_utilities.scss
```diff
@@ -1,5 +1,11 @@
 @import "~nhsuk-frontend/packages/core/settings/colours";
 // re-usable styles and mixins
+
+h2 {
```
hihilary, line 4: The `h2` tag is used somewhere else (eg. notes-drawer, confirm-recommendation, etc.), and the `notes-drawer.component.scss` also has the same h2 style. Should we remove that one?

hihilary (approved):

PR #937, opened 2022-07-07
Feat/rename connection tabs

StevenHee:
https://hee-tis.atlassian.net/browse/TIS21-3060


CaiWillis:
LGTM

Assuming if we propagate this rename to the BE + endpoints (more tech improvement than direct user value) then any FE work can be subtasks of those tickets

StevenHee:
> LGTM
> 
> Assuming if we propagate this rename to the BE + endpoints (more tech improvement than direct user value) then any FE work can be subtasks of those tickets

Yes, absolutely. The ticket is already created and can add as a subtask.
https://hee-tis.atlassian.net/browse/TIS21-3175

PR #947, opened 2022-08-04
Feat/deferal validation

StevenHee:
When the submission due date is greater than 120 days from today, the defer option in the recommendation dropdown is disabled and a warning message is displayed advising the user why this is disabled and the date from when deferral is possible.

Inline form validation messages have also been enabled as this is better UX than disabling save/submit without any explanation.

TICKET: https://hee-tis.atlassian.net/browse/TIS21-3241


jayanta2018:
> m validation messages have also been enabled as this is better UX than disabling save/submit without any explanation.

jayanta2018 (approved):
LGTM

PR #976, opened 2022-10-17
Feat/gmc status

StevenHee:
TICKET: https://hee-tis.atlassian.net/browse/TIS21-3555

### DIFF
--- a/cypress/e2e/app/recommendations/filter-summary-records.cy.ts
+++ b/cypress/e2e/app/recommendations/filter-summary-records.cy.ts
@@ -229,7 +229,7 @@ describe("Recommendations", () => {
       });
     });
 
-    describe.only("Filter by Programme Name", () => {
+    describe("Filter by Programme Name", () => {
       it("should display 'programme name' filter field", () => {
         initFilterPanel();
         cy.get("[data-cy='formfield_programmeName']").should("exist");
--- a/cypress/e2e/app/recommendations/filter-summary-records.cy.ts
+++ b/cypress/e2e/app/recommendations/filter-summary-records.cy.ts
@@ -3,11 +3,18 @@ describe("Recommendations", () => {
     beforeEach(() => {
       cy.loginSession();
     });
+
+    const openProgrammeNameDropdown = (query: string = "General") => {
+      cy.get("mat-option").should("not.exist");
+      cy.get("[data-cy='formfield_programmeName'] input").type(query);
+      cy.wait(2000);
+      cy.get("mat-option").should("exist");
+    };
+
     const initFilterPanel = () => {
       cy.visit("/recommendations");
       cy.get("[data-cy='toggleTableFiltersButton']").click();
       cy.get(".filters-drawer-container .mat-drawer-opened").should("exist");
-      cy.wait(3000);
     };
 
     const checkButtonsDisabled = (status: string = "be disabled") => {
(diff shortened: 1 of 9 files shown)


cypress/e2e/app/recommendations/filter-summary-records.cy.ts
```diff
@@ -155,12 +155,80 @@ describe("Recommendations", () => {
...
+
+        cy.get(
+          "[data-cy='selectionListgmcStatus'] .mat-list-option .mat-list-text"
+        ).should("have.length", 3);
+        cy.get(
+          "[data-cy='selectionListgmcStatus'] .mat-list-option .mat-list-text"
+        ).each(($el) => {
+          expect(gmcStatusLabels.indexOf($el.text().trim())).to.be.greaterThan(
+            -1
+          );
```
JosephKelly, line 183: This is cleaner as `expect($el.text().trim()).to.be.oneOf(gmcStatusLabels);`

cypress/e2e/app/recommendations/filter-summary-records.cy.ts
```diff
@@ -155,12 +155,80 @@ describe("Recommendations", () => {
...
     });
 
+    describe("Filter by GMC status", () => {
+      it("should display 'gmc status' selection list filters", () => {
+        initFilterPanel();
+        cy.get("[data-cy='selectionListgmcStatus']").should("exist");
+      });
+
+      it("should display 3 selection options in the 'gmc status' selection list labelled accordingly", () => {
+        const gmcStatusLabels = ["Approved", "Rejected", "Under Review"];
```
JosephKelly, line 171: Looking at where some or all of these are referenced, would it be worth defining these elsewhere or using values from `src/app/recommendations/constants.ts`?

cypress/e2e/app/recommendations/filter-summary-records.cy.ts
```diff
@@ -155,12 +155,80 @@ describe("Recommendations", () => {
...
+          .then(() => {
+            expect(
+              gmcStatus.every((status) => {
+                return (
+                  status === selectedGmcStatus[0] ||
+                  status === selectedGmcStatus[1]
+                );
+              })
+            ).to.be.true;
+          });
```
JosephKelly, line 228: This seems like a different way of doing `expect($el.text().trim()).to.be.oneOf(selectedGmcStatus);` in another test.  If so, I expect it should be implemented in the same way.

JosephKelly (changes requested):
1 question and a couple of things that I think should change

cypress/e2e/app/recommendations/filter-summary-records.cy.ts
StevenHee: This is an interesting point. To what extent should your test code be independent from your code? As a tester I'm defining what I expect to see which I feel should be separate to what the programmer has defined.

JosephKelly: Yes.  We definitely want it to be independent of implementation specifics and we stop short of having a different git repo for tests.  An assumption: the test dependencies include the "implementation" dependencies too.  Where we have constants, they operate a lot like that "common" config.
This was the question anyhow.  Test isolation, as in tests that are independent of each other and therefore don't cause each other to break might also be a reason you want to define this multiple times across tests but I took these labels as "`const`ant" enough that if they change in one place, we'd need to change them everywhere.
Happy for you to resolve & continue conversation elsewhere if that's useful.

StevenHee:
I have refactored the E2E to use the `to.be.oneOf` assertion in `GMC status` and `dbc` tests. I have attempted to remove any unnecessary `cy.wait` commands, which from testing are still required in the programme name drop down and form submission. As for defining constants elsewhere, there doesn't appear to be an obvious or clean way of doing this in Cypress so I suggest a bit more research with possible implementation as part of next filtering ticket https://hee-tis.atlassian.net/browse/TIS21-3603

JosephKelly (approved):
nice refactor of that common task

PR #977, opened 2022-10-20
test:fix to.be.oneOf assertion

StevenHee:
I cocked up my initial refactoring of the `to.be.oneOf` assertion in E2E tests.


JosephKelly (approved):
:facepalm: I can't believe I didn't read it properly against my comment on the last PR!

PR #983, opened 2022-11-03
Feat/filter by tis admin

StevenHee:
https://hee-tis.atlassian.net/browse/TIS21-3742

### DIFF
--- a/src/app/recommendations/recommendations.resolver.ts
+++ b/src/app/recommendations/recommendations.resolver.ts
@@ -62,7 +62,7 @@ export class RecommendationsResolver
       )
       .subscribe((admins: IAdmin[]) => {
         const tisAdminFormField = TABLE_FILTERS_FORM_BASE.find(
-          ({ key }) => key === "tisAdmin"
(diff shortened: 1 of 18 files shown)


src/app/recommendations/recommendations.resolver.ts
```diff
@@ -34,6 +40,33 @@ export class RecommendationsResolver
...
   }
 
+  admins$: Observable<IAdmin[]> = this.store.select(
+    (state) => state.admins.items
+  );
+
+  private async initFiltersFormData() {
+    this.admins$
+      .pipe(
+        filter((data) => !!data),
```
CaiWillis, line 50: Is this a check for "not undefined"?

StevenHee: Yes. The observable gets subscribed to twice, the first time it's undefined. Instead of checking the data is returned from the subscription, I filter it out in the pipe before, such that it only subscribes when there is data.

CaiWillis (approved):

PR #988, opened 2022-11-09
Feat/restyle multiple programmes

StevenHee:
A few tweaks to FE in this PR.

- When a doctor has multiple programmes, the data returned includes each programme name separated by a `|`. This is processed in the recommendations summary table and doctors detail side nav via a new pipe that outputs each programme name enclosed in a `<span>` tag to allow displaying on separate lines and custom styling.
- The mat list used in the details side nav restricts content to a single line, which was truncating longer values such as programme name. This is fixed via css changes.
- The Programme name label is updated to Current programme name.
- Bug fix whereby the data in recommendations summary table was sorted by GMC number when page is reset and query string params are present with sort column param set to submission due date. 
- Fix E2E tests

### DIFF
--- a/src/app/details/details-side-nav/details-side-nav.component.html
+++ b/src/app/details/details-side-nav/details-side-nav.component.html
@@ -14,12 +14,18 @@
       <div class="mat-caption" mat-line>GMC No:</div>
       <div mat-line>{{ traineeDetails.gmcNumber }}</div>
     </mat-list-item>
-    <mat-list-item *ngIf="traineeDetails.programmeMembershipType">
+    <mat-list-item>
       <div class="mat-caption" mat-line>Programme membership type:</div>
-      <div mat-line>{{ traineeDetails.programmeMembershipType }}</div>
+      <div mat-line>
+        {{
+          traineeDetails.programmeMembershipType === null
+            ? "N/A"
+            : traineeDetails.programmeMembershipType
+        }}
+      </div>
     </mat-list-item>
     <mat-list-item *ngIf="traineeDetails.programmeName">
-      <div class="mat-caption" mat-line>Programme name:</div>
+      <div class="mat-caption" mat-line>Current programme name:</div>
       <div
         mat-line
         class="programme-names"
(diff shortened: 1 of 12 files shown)


CaiWillis (approved):

src/app/details/details-side-nav/details-side-nav.component.html
```diff
@@ -14,14 +14,25 @@
...
     </mat-list-item>
     <mat-list-item *ngIf="traineeDetails.programmeName">
-      <div class="mat-caption" mat-line>Programme name:</div>
-      <div mat-line>{{ traineeDetails.programmeName }}</div>
+      <div class="mat-caption" mat-line>Current programme name:</div>
+      <div
+        mat-line
+        class="programme-names"
+        [innerHTML]="traineeDetails.programmeName | splitStringToHTML"
+      ></div>
```
hihilary, line 33: As Pepe spotted, we could still need current programme name label even if it's empty.

src/app/shared/pipes/split-string-to-html.pipe.ts
```diff
@@ -0,0 +1,16 @@
...
+@Pipe({
+  name: "splitStringToHTML"
+})
+export class SplitStringToHTMLPipe implements PipeTransform {
+  transform(value: string, delimiter: string = "|"): SafeHtml {
+    if (value && value.includes(delimiter)) {
+      return value
+        .split(delimiter)
+        .map((item: string) => `<span class='split-item'>${item}</span>`)
+        .join(" ");
```
hihilary, line 12: The delimiter from backend actually is " | " which contains a leading and a trailing whitespace, I'm a little wondering if we need to trim them.

StevenHee: The output will have redundant space  `foo | bar => <span>foo </span><span> bar</span>` From a presentation perspective this extra whitespace has no impact as it is ignored when rendered in browser. Alternatively, the pipe can take any delimiter so I can pass ` | ` instead.

src/app/details/details-side-nav/details-side-nav.component.html
StevenHee: Fixed

hihilary (approved):

PR #989, opened 2022-11-09
feat: remove connections actions from UI

CaiWillis:
TIS-3732: Remove interactions

### DIFF
--- a/cypress/e2e/app/connections/connection-component/connection-component.cy.ts
+++ b/cypress/e2e/app/connections/connection-component/connection-component.cy.ts
@@ -1,0 +1,15 @@
+describe("Connections Connection Component", () => {
+    beforeEach(() => {
+      cy.login();
+    });
+  
+    const button = "app-update-connetions-btn";
+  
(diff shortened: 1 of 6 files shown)


CaiWillis:
- Decided that editing what roles can or can't do would be counter productive/innaccurate

- Went for the brute force approach of just hiding/removing the components

- Attempted some cypress tests but then realised that I would have to nevigate the fact that when you visit the page, it breaks automatically. Still worth doing?

cypress/e2e/app/connections/connection-component/connection-component.cy.ts
```diff
@@ -0,0 +1,20 @@
+describe("Connections Connection Component", () => {
+    beforeEach(() => {
+      cy.login();
+    });
```
StevenHee, line 4: Using `cy.login` method requires login between each test. Unless you specifically want to force login for each test, it may be quicker to use `cy.loginSession` instead.

StevenHee (approved):

PR #1003, opened 2022-11-30
Fix/security issues

StevenHee:
Having addressed some of the security issues via dependabot PRs, others are better fixed by updating root packages where possible. Upgrading to use the latest version of node v18 provides access to the overrides feature allowing patching of individual child packages.

NO TICKET


JosephKelly (approved):
I like the thoughtfulness in the docs.

PR #1028, opened 2023-01-16
fix/hotjarSiteId

JosephKelly:
The value was for the preprod environment.
There may well be additional config but this is the site id from hotjar.

NO-CARD: JFDI fix


JosephKelly:
Thought I'd add a note that we are currently getting the recordings, just under the "stage" hotjar site.

StevenHee (approved):

PR #1029, opened 2023-01-19
chore: move deployments into single workflow

JosephKelly:
It doesn't need to be separated anymore.
There is the same run ordering issue present everywhere else.

Deleted step to push a stable tag.
It ran before E2E tests so isn't necessarily stable.

NO-CARD: Suggestion following deploy comment


Judge40 (approved):
Approving as a devops representative, up to you if you want an internal reviewer too.

PR #1045, opened 2023-02-07
feat:add alpha env variables for callback

StevenHee:
Need to set sign/out domain for alpha. Conditional test for alpha domain temporary so detection via `window.location` should suffice.

### DIFF
--- a/src/assets/environments/environment.json
+++ b/src/assets/environments/environment.json
@@ -45,5 +45,28 @@
       "redirectSignIn": "https://stage-revalidation.tis.nhs.uk",
       "redirectSignOut": "https://stage-revalidation.tis.nhs.uk"
     }
+  },
+  "alpha": {
+    "production": true,
+    "name": "stage",
+    "siteIds": ["UA-40570867-6"],
+    "hotJarId": 1980924,
+    "hotJarSv": 6,
+    "adminsUIHostUri": "https://stage-apps.tis.nhs.uk/",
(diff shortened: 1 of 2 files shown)


src/assets/environments/environment.json
```diff
@@ -45,5 +45,28 @@
       "redirectSignIn": "https://stage-revalidation.tis.nhs.uk",
       "redirectSignOut": "https://stage-revalidation.tis.nhs.uk"
     }
+  },
+  "alpha": {
+    "production": true,
```
CaiWillis, line 50: Everything else here seems to be stage/preprod, is this correct?

StevenHee: > Everything else here seems to be stage/preprod, is this correct?

Yes, that's correct. Alpha is using all the backend services from preprod including cognito. The only real differences is once its authenticated, it returns to alpha FE

src/assets/environments/environment.json
```diff
@@ -45,5 +45,28 @@
...
+    "dateFormat": "dd/MM/yyyy",
+    "awsConfig": {
+      "domain": "stage-auth.tis.nhs.uk",
+      "region": "eu-west-2",
+      "scope": ["openid", "aws.cognito.signin.user.admin"],
+      "authenticationFlowType": "USER_PASSWORD_AUTH",
+      "responseType": "token",
+      "userPoolId": "eu-west-2_o5Es6Bpl8",
+      "userPoolWebClientId": "r28c0u6jpc7scm9e2anhh274r",
+      "bucketName": "tis-revalidation-concerns-upload-preprod",
```
CaiWillis, line 66: wrong bucket name?

StevenHee: > wrong bucket name?

It's fine. The only distinction between alpha and preprod is the hosting of the FE

CaiWillis (approved):

PR #1051, opened 2023-02-20
feat:add deploy to s3 bucket job in pipeline

StevenHee:


JosephKelly (approved):
Approved as a backward compatible step toward switching to S3 hosting

PR #1056, opened 2023-02-27
feat:update prod deployment to s3

StevenHee:
- Update the deploy-to-prod job to use s3
- Remove ECS jobs - build, build-artifacts and dockerise
- Change AWS_S3_BUCKET and CLOUDFRONT_DISTRIBUTION_ID secrets to vars

TIS21-4043: CI/CD pipeline modification to deploy to S3 in parallel with existing pipeline

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -41,10 +41,10 @@ jobs:
         run: npm run test:ci
 
       - name: Deploy app build to S3 bucket
-        run: aws s3 sync ./dist/revalidation/ s3://${{ secrets.AWS_S3_BUCKET }} --delete
+        run: aws s3 sync ./dist/revalidation/ s3://${{ vars.AWS_S3_BUCKET }} --delete
 
       - name: Invalidate CDN Cache
-        run: aws cloudfront create-invalidation --distribution-id ${{ secrets.CLOUDFRONT_DISTRIBUTION_ID}} --paths /*.js /*.css /static/*
+        run: aws cloudfront create-invalidation --distribution-id ${{ vars.CLOUDFRONT_DISTRIBUTION_ID}} --paths /*.js /*.css /static/*
 
       - name: Slack Notification - Deploy Success
         if: success()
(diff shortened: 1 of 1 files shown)


Judge40 (approved):
Looks good to this trainee dev

.github/workflows/deploy.yml
```diff
@@ -207,53 +115,57 @@ jobs:
...
 
-      - name: Fill in the new image ID in the Amazon ECS task definition
-        id: task-def
-        uses: aws-actions/amazon-ecs-render-task-definition@v1
-        with:
-          task-definition: .aws/task-definition-prod.json
-          container-name: reval-fe
-          image: ${{ steps.login-ecr.outputs.registry }}/tis-revalidation-v2:${{ github.sha }}
+      - name: Build Angular App
+        run: npm run build:misc && npm run build -- --configuration=production
```
JosephKelly, line 143: There's an opportunity here to remove what looks like a redundant step.  This duplicates the step on L#37.  Approved as I believe there is something in progress to look at the workflow but thought I'd flag it as something to address if that won't be applied here soon

JosephKelly (approved):

.github/workflows/deploy.yml
Judge40: So we'd upload the preprod files as an artifact on the workflow run and then retrieve them on prod deploy?
We only store artifacts for 7 days, workflow approvals will last 30 days currently so there's a big window in which the prod deploy is a guaranteed fail. Syncing between preprod and prod buckets is asking for trouble, so I'm not sure where else to store the preprod build.

This might actually be a good opportunity to reconsider the requirement for the `/api/environment` endpoint, we reworked TSS when we moved to S3 to also set the env vars at build time instead of runtime.
https://github.com/Health-Education-England/tis-trainee-ui/blob/main/package.json#L77
https://github.com/Health-Education-England/tis-trainee-ui/blob/main/build.sh#L4
https://github.com/Health-Education-England/tis-trainee-ui/blob/main/.github/workflows/deploy.yml#L35

JosephKelly: Yeah & it would be good to keep the workflows semantically in step.
Hopefully GitHub'll "fix" that 30 day thing.  Leaving a deployment around for that long seems to be opening deployments up for lots of "fun".

PR #1062, opened 2023-03-08
feat: change authflow from token to code.

StevenHee:
Implicit token flow does not support refresh tokens meaning the users were forced tologin every 5 minutes. Changing to use auth code flow enables refresh tokens.

TIS21-4297: Update Reval to use authorization code flow with Cognito to support refresh tokens


Judge40 (approved):
Should alpha be updated too? otherwise LGT-A-TSS-DEV

StevenHee:
> Should alpha be updated too? otherwise LGT-A-TSS-DEV

I think plan is to bin the alpha environment at some point so didn't see the need.

PR #1063, opened 2023-03-10
Refactor/env variables

StevenHee:
Following the recent changes to hosting of Reval website in S3, the code can now be built and deployed separately for `stage` and `prod` in the pipeline. This means the application can utilise the built-in Angular approach to handling environment variables, which this PR attempts to implement. https://angular.io/guide/build. I have raised as a draft PR in case I have overlooked an obvious reason not to adopt the Angular approach.


JosephKelly (approved):
As far as I can tell it looks ok.
Relying on you to decide that a build per environment doesn't work too much against knowing what is deployed in stage with behave in the same way on prod.

PR #1121, opened 2023-06-12
Feat/filter connections by programme

StevenHee:
I have also changed the bulk update connection icon and tooltip in this PR following feedback in UR sessions.

TIS21-4635: Reval Connections MVP: Filtering by programme name


CaiWillis:
I'm kind of okay with that duplication, you _could_ refactor it out into a generic "filters tests" folder but then that makes the test structure a bit more confusing, what do you think?

CaiWillis (approved):

PR #1124, opened 2023-06-16
wip: use different aggragation for sidebar nav details

CaiWillis:


CaiWillis:
Not needed, worked out a better backend solution

PR #1129, opened 2023-06-20
test:update dbcs to use abbreviations in E2E test

StevenHee:
TIS21-4655: Display the DBC as an abbreviation instead of code in summary records table


CaiWillis (approved):
Tidy

PR #1160, opened 2023-07-31
Chore/upgrade to angular16

StevenHee:
This upgrades Angular to v16 since v13 is no longer supported. If follows the incremental upgrade steps outlined in https://update.angular.io/, hence the number of commits. Other packages have also been updated in line with this. In addition to updating `package.json`, as of Angular 14, reactive forms are strictly typed by default (https://angular.io/guide/typed-forms). Any component using forms have been automatically updated to import the untyped class instead, hence the number of files updated. Also, the 2 packages used for HotJar and Google Analytics weren't compatible with v16. Instead, I have switched to use Google Tag Manager. This allows tags like GA and HotJar to be added via the GTM UI rather than code. The updates have been deployed to https://alpha-revalidation.tis.nhs.uk for testing and it is capturing analytics and HotJar events successfully.

There isn't a ticket for this as it is a chore that addresses some the dependabot updates that couldn't be patched individually. The only enhancement has been the shift to GTM, which is another way of deploying what we already have. 

NO TICKET


src/app/app.module.ts
```diff
@@ -25,10 +25,11 @@ import { initializeApplication } from "./core/auth/auth.initializer";
 import { errorHandlerFactory } from "./factories/error-handler.factory";
 
 import { swRegistrationOptionsFactory } from "./factories/sw-registration-options.factory";
-import { NgxHotjarModule } from "ngx-hotjar";
```
CaiWillis, line 28: Just checking these removals of hotjar are intended

CaiWillis: Just re-read the description comment

CaiWillis (approved):

src/app/app.module.ts
StevenHee: Yep. These 3rd party modules are not compatible with v16 and haven't been updated for some time. The new approach is to use Google Tag Manager, and then add HotJar and GA4 (and any other tags we want in future) from there. This then avoids the need for adding separate tags into the code.

PR #1174, opened 2023-08-22
Fix/bulk update connections checkboxes

StevenHee:
The checkboxes would not appear when programme name contained Foundation or membership type was Military. Since F1 and Military are excluded in data and the All Doctors tab is no longer required, these conditionals have been removed.

TICKET: TIS21-4928 - Bulk Connections seems to break when list is filtered by foundation (F2) programmes


jayanta2018 (approved):
Ran succefully at my local with node version 16.14.0. LGTM

PR #1183, opened 2023-08-30
fix: return both data fetches once complete

StevenHee:
Due to the intermittent delay in fetching data from Trainee endpoint on Prod, the previous fix wasn't sufficient. CombineLatest waits for items to be emitted from both observables before returning and resolving to the details page.

The browser won't navigate to the details page until data is returned so when there is any significant delay, there is no indication in the UI that the details page is loading. Either the performance issues will need addressing or some form of loading spinner could be added.

TIS21-4868 - Prevent details page from initially displaying previous loaded doctor's details in the trainee profile panel

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "revalidation",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "scripts": {
     "ng": "ng",
     "start": "npm run build:misc && ng serve",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "revalidation",
-  "version": "1.0.0",
+  "version": "1.0.1",
```
hihilary, line 3: 😄

hihilary (approved):

PR #1210, opened 2023-10-20
Feat/exceptions log

StevenHee:
Display the exceptions log for connection updates made by the current user today. The records list component isn't suitable to display the exceptions due to the different data models and features like pagination, search and filtering are not required. However, it feels appropriate to add the link as a new tab under connections. A new component and service is used to fetch and display the exceptions log which is conditionally displayed in the connections section according to the filter property.

TIS21-5151: Show log of errors when adding/removing connections

### DIFF
--- a/src/app/connections/exceptions-log/exceptions-log.component.html
+++ b/src/app/connections/exceptions-log/exceptions-log.component.html
@@ -1,0 +1,66 @@
+<ng-container id="ShowError" *ngIf="pageStatus === 'error'">
+  <div data-test="ErrorMessage" class="mt-20 d-flex justify-content-center">
+    Exceptions data failed to load.
+  </div>
+</ng-container>
+<ng-container id="ShowLoadingSpinner" *ngIf="pageStatus === 'loading'">
+  <div
+    class="d-flex align-content-center justify-content-center align-items-center progress-spinner"
+  >
+    <mat-spinner></mat-spinner>
+  </div>
+</ng-container>
+
+<ng-container id="NoExceptionDataFound" *ngIf="pageStatus === 'empty'">
+  <div
+    data-test="NoExceptionsMessage"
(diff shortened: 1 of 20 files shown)


src/app/connections/exceptions-log/exceptions-log.component.html
```diff
@@ -0,0 +1,66 @@
...
+    data-test="NoExceptionsMessage"
+    class="mt-20 d-flex justify-content-center"
+  >
+    No exceptions found.
+  </div>
+</ng-container>
+
+<ng-container id="ExceptionDataFound" *ngIf="pageStatus === 'ok'">
+  <h2 class="ml-20 mt-20">
+    List of errors from unsuccessful connection updates made today by currennt
```
CaiWillis, line 25: "current"

CaiWillis: Also just "unsuccessful connections" sounds smoother I think (I know technically these are for disconnections too but . . .)

src/app/connections/exceptions-log/mock-data/exceptions-log-spec-data.ts
```diff
@@ -0,0 +1,27 @@
+import { IException } from "../exceptions-log.interface";
+
+export const exceptionsJson = `
+[
+   
+    {
+        "gmcId": "1",
+        "errorMessage": "Error message",
+        "timestamp": "2023-10-18T14:45:47.337",
```
CaiWillis, line 9: Will these times work in future? (I'm guessing the "flush" method used just returns this no matter what day it is but just checking)

StevenHee: Yeah, this uses the test data so should always pass.

CaiWillis (approved):

PR #1239, opened 2023-12-12
fix: temporary patch for recommendations error

CaiWillis:
TIS21-5504: Unscheduled full production data-reynchronisation accidentally triggered


CaiWillis:
better fix from FE developer

PR #1390, opened 2025-09-03
fix: set dbc params to default if no filters selected for discrepancies

CaiWillis:
TIS21-7614: Reval Discrepancies: Filtering by programme

### DIFF
--- a/src/app/connections/services/connections.service.ts
+++ b/src/app/connections/services/connections.service.ts
@@ -24,16 +24,17 @@ export class ConnectionsService {
       .append("programmeName", snapshot.tableFilters?.programmeName || "");
 
     if (snapshot.filter === "discrepancies") {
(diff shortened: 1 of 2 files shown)


src/app/connections/services/connections.service.ts
```diff
@@ -24,19 +24,21 @@ export class ConnectionsService {
       .append("programmeName", snapshot.tableFilters?.programmeName || "");
 
     if (snapshot.filter === "discrepancies") {
+      let noneSelected = !(snapshot.tableFilters?.dbcs || snapshot.tableFilters?.tisDesignatedBodies)
```
CaiWillis, line 27: I realise I could drop the ! on both of these for simpler logic, but I went for what I thought was the clearer naming & intent option

Let me know if "atLeastOneSelected" is preferable

PR #1391, opened 2025-09-06
feat: add TOTP to E2E login command

StevenHee:
Add cyress-otp package
Add generateOtp as task
Update login command to handle OTP
Update deploy GHA to fetch totp key from secrets
Update Cypress to fix critical dependabot issue

TIS21-7621

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "revalidation",
-  "version": "1.2.1",
+  "version": "1.2.3",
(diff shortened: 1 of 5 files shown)


package-lock.json
CaiWillis, line 1: Was this included deliberately?

StevenHee: Yeah. I understand it's good practice to keep this updated in the repo.

JosephKelly (approved):

PR #1421, opened 2026-04-13
fix: update axios

StevenHee:
A critical security bug is found in axios version used by some v16 angular packages. Overriding the version to resolve.

Updating aws-amplify to latest to resolve critical security bug

NO TICKET

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "revalidation",
-  "version": "1.8.5",
+  "version": "1.8.6",
   "scripts": {
     "ng": "ng",
     "start": "npm run build:misc && ng serve",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -96,5 +96,8 @@
     "tslint": "~6.1.3",
     "tslint-config-prettier": "^1.18.0",
     "typescript": "^5.1.6"
+  },
+   "overrides": {
+    "axios": "^1.15.0"
```
JosephKelly, line 101: I'm noting the difference from the suggested version to pin against.  As the suggestion is later and there's no notification related this version, it seems sensible enough to me; moving from one unaffected version to a later unaffected version.

JosephKelly (approved):

PR #1422, opened 2026-04-13
chore/upgrade angular v17

StevenHee:
The upgrade of Angular to version 17 in Reval is almost seamless, requiring only minimal changes to the project files. Included are upgrading certain additional packages and removing some that are redundant. 

NO TICKET


src/scss/material/_index.scss
```diff
@@ -33,8 +33,6 @@ $nhsuk-theme: mat.define-light-theme(
 );
 
 // Legacy - only needed whilst migrating to MDC
```
CaiWillis: Remove comment

StevenHee: Removed

CaiWillis (approved):

